PR #3, opened 2018-05-17
Added scripts to enable root user

iamshital:
This will unblock performance tests.

### DIFF
--- a/Testscripts/Linux/enablePasswordLessRoot.sh
+++ b/Testscripts/Linux/enablePasswordLessRoot.sh
@@ -1,0 +1,25 @@
+#!/bin/bash
+#AUTHOR : SHITAL SAVEKAR <v-shisav@microsoft.com>
+#Description : Enables passwordless authentication for root user.
+#How to use : ./enablePasswordLessRoot.sh
+#In multi VM cluster. Execute this script in one VM. It will create a sshFix.tar
+#Copy this sshFix.tar to other VMs (/root) in your cluster and execute same script. It will extract previously created keys.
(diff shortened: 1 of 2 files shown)


Testscripts/Linux/enablePasswordLessRoot.sh
```diff
@@ -0,0 +1,25 @@
+#!/bin/bash
+#AUTHOR : SHITAL SAVEKAR <v-shisav@microsoft.com>
+#Description : Enables passwordless authentication for root user.
+#How to use : ./enablePasswordLessRoot.sh
+#In multi VM cluster. Execute this script in one VM. It will create a sshFix.tar
+#Copy this sshFix.tar to other VMs (/root) in your cluster and execute same script. It will extract previously created keys.
+#This way, all VMs will have same public and private keys in .ssh folder.
+
+rm -rf /root/.ssh
```
nicepat: can you add conditional statement for this 'rm'?

Testscripts/Linux/enablePasswordLessRoot.sh
```diff
@@ -0,0 +1,25 @@
...
+cd /root
+keyTarFile=sshFix.tar
```
nicepat: Please add a conditional statement is file existing

Testscripts/Linux/enablePasswordLessRoot.sh
iamshital: We're forcibly removing the .ssh folder because, everytime, automation creates new keys with the same name. If the keys are present there, then it might be problem executing the test. So this is safe.

PR #13, opened 2018-05-19
Added SRIOV network performance tests

iamshital:


iamshital:
This tag is not-necessary for this particular test and planned to be removed. 
If we delete it immediately, the automation will break. We're planning to keep only used tags to XML elements.

PR #16, opened 2018-05-23
XFS Tests + Scripts required to support CustomVHD testing.

iamshital:


nicepat:
Utilities/CopyVHDtoOtherStorageAccount.ps1 has no lines for catch and finally part. Please add reasonable instructions.

iamshital:
@nicepat , added code to try/catch/finally block

PR #20, opened 2018-05-23
Changed the area of XFS tests

iamshital:


nicepat:
This is only for "TestByCategorisedTestname" not "TestByCategory" Does this change apply to both menu? Also what about "tag"?

iamshital:
Its updated everywhere.
image

iamshital:
tags are relevant. CIFS test uses smb3 protocol.
Ext4 and Xfs tests are ran directly on the disk. Hence they are disk related.

PR #37, opened 2018-05-30
Added support to pass parameters using parameters file as well as command line

iamshital:


Utilities/LaunchTestPipelineRemotely.sh
```diff
@@ -1,57 +1,35 @@
...
-LinuxPassword=""
-
-#Required to access Jenkins.
-ApiToken=""
-#Required to upload Files to jenkins server using FTP
-FtpUsername=""
-FtpPassword=""
-
-#Required
-JenkinsURL="penguinator.westus2.cloudapp.azure.com"
```
nicepat: Should be hardcoded

Utilities/LaunchTestPipelineRemotely.sh
```diff
@@ -1,57 +1,35 @@
...
-
-#Required
-#TestPipeline="/job/Microsoft-Test-Execution-Pipeline"
-TestPipeline=""
-
-
+#####################################################################################################
+#How to use:
```
nicepat: At the end of script execution line, it should have BlueOcean URL

Utilities/LaunchTestPipelineRemotely.sh
```diff
@@ -1,57 +1,35 @@
 #!/bin/bash
 
-#Required
```
nicepat: Add friendly comment - what it is each in comment.

PR #89, opened 2018-07-02
Added single connection TCP performance test (iperf3)

iamshital:


Libraries/HyperV.psm1
```diff
@@ -354,9 +354,15 @@ Function CreateHyperVGroupDeployment([string]$HyperVGroup, $HyperVGroupNameXML)
                 if ($?)
                 {
                     LogMsg "Set-VM -VM $($NewVM.Name) -ProcessorCount $CurrentVMCpu -StaticMemory -CheckpointType Disabled -Notes $HyperVGroupName"
+
                     $Out = Set-VM -VM $NewVM -ProcessorCount $CurrentVMCpu -StaticMemory  -CheckpointType Disabled -Notes "$HyperVGroupName"
                     LogMsg "Add-VMGroupMember -Name $HyperVGroupName -VM $($NewVM.Name)"
                     $Out = Add-VMGroupMember -Name "$HyperVGroupName" -VM $NewVM -ComputerName $HyperVHost
+                    $ResourceDiskPath = ".\Temp\ResourceDisk-$((Get-Date).Ticks)-sdb.vhd"
```
juhlee-microsoft: Test-Path?

Testscripts/Windows/PERF-NETWORK-TCP-SINGLE-CONNECTION-THROUGHPUT.ps1
```diff
@@ -0,0 +1,296 @@
...
+		LogMsg "  SSH Port  : $($serverVMData.SSHPort)"
+		LogMsg "  Location  : $($serverVMData.Location)"
+		#
+		# PROVISION VMS FOR LISA WILL ENABLE ROOT USER AND WILL MAKE ENABLE PASSWORDLESS AUTHENTICATION ACROSS ALL VMS IN SAME HOSTED SERVICE.	
+		#
+		ProvisionVMsForLisa -allVMData $allVMData -installPackagesOnRoleNames "none"
+
+		#endregion
+
+		if ($EnableAcceleratedNetworking -or $CurrentTestData.AdditionalHWConfig.Networking -imatch "SRIOV")
```
juhlee-microsoft: checking both conditions in true? I think this should be -and. ?

Testscripts/Windows/PERF-NETWORK-TCP-SINGLE-CONNECTION-THROUGHPUT.ps1
```diff
@@ -0,0 +1,296 @@
...
+			Add-Content -Value "client=$($clientVMData.InternalIP)" -Path $constantsFile
+		}
+		else
+		{
+			Add-Content -Value "server=$($serverVMData.PublicIP)" -Path $constantsFile	
+			Add-Content -Value "client=$($clientVMData.PublicIP)" -Path $constantsFile
+		}
+		#endregion
+
+		foreach ( $param in $currentTestData.TestParameters.param)
```
juhlee-microsoft: Inside foreach loop, consider init value of testBuffers, testConnections, IPVersion in case they could not get through in truth. It will be CommandNotFoundException

Testscripts/Windows/PERF-NETWORK-TCP-SINGLE-CONNECTION-THROUGHPUT.ps1
iamshital: OR condition is correct here.
Either we can enable accelerated networking by mentioned into the Test Parameters or we can enable it by mentioning into VM Configuration

Testscripts/Windows/PERF-NETWORK-TCP-SINGLE-CONNECTION-THROUGHPUT.ps1
iamshital: test buffers, test connections values will be mentioned in test case definition. so they will be always present.

Testscripts/Windows/PERF-NETWORK-TCP-SINGLE-CONNECTION-THROUGHPUT.ps1
juhlee-microsoft: acked

PR #100, opened 2018-07-17
Add case NESTED-KVM-NTTTCP-PUBLIC-BRIDGE.

LiliDeng:
Support test run against 2 servers. Support add one more nic.
Specify server names use this parameter -TestLocation 'name1,name2'.


LiliDeng:
Will submit a new PR, sorry for confuse

PR #103, opened 2018-07-20
Update with some cosmetic fix

simonxiaoss:


Libraries/CommonFunctions.psm1
```diff
@@ -26,12 +26,22 @@
 
 Function ThrowException($Exception)
 {
```
simonxiaoss: There are two functions implemented with the same name "ThrowException". One is here and another one is from "Framework.psm1". 
Merge into one.

RunTests.ps1
```diff
@@ -33,6 +33,7 @@ Param(
     $BuildNumber=$env:BUILD_NUMBER,
 
     #[Required]
+    [ValidateSet('Azure','HyperV')]
```
simonxiaoss: PS will validate the range of the value automatically for "ValidateSet" parameters.

PR #112, opened 2018-07-27
change the ValiateXMLs call

simonxiaoss:
It makes sense to only validate the xml files once, after all XML files are prepared.


lizzha:
@simonxiaoss could you please resolve the conflicts? Then we will merge the PR

Libraries/Framework.psm1
```diff
@@ -187,7 +187,6 @@ Function UpdateGlobalConfigurationXML()
 		}
 		if ( $XMLSecretFile )
 		{
-			ValiateXMLs -ParentFolder $((Get-Item -Path $XMLSecretFile).FullName | Split-Path -Parent)
```
ader1990: Can you also change the typo in this method name?

simonxiaoss:
I will rebase the code and submit a new PR (also fixing the method name typo).

PR #113, opened 2018-07-27
Juhlee/add blob access

juhlee-microsoft:
This is for Jenkins execution. It will access Azure blob location if blob location tag is passed in AzureSecret.xml file


juhlee-microsoft:
Local test and Jenkins test done. Jenkins test itself is passed, but uploading the result failed. Do not know if this comes from lab move or not.

RunTests.ps1
```diff
@@ -93,13 +93,24 @@ try {
...
 		# Verify the binary file in Tools location
 		if ( Test-Path $CurrentDirectory/Tools/$_ ) {
 			Write-Host "$_ File exists already and available to use in Tools folder."
 		} else {
-			Write-Error "File not found in Tools folder: $_. Testing terminates."
-			throw [System.IO.FileNotFoundException]
+
+			$WebClient.DownloadFile("$azureBlobLoc/$_","$CurrentDirectory\Tools\$_")
+
+			if (Test-Path "$CurrentDirectory\Tools\$_") {
```
lizzha: If the file is not downloaded successfully, will the call $WebClient.DownloadFile throw exception directly? If yes, then there's no need to Test-Path for the local file.

.vscode/settings.json
```diff
@@ -0,0 +1,2 @@
+{
+}
```
ader1990: can you remove this file, it has nothing to do with the purpose of the commit

RunTests.ps1
```diff
@@ -93,13 +93,24 @@ try {
 	$CurrentDirectory = Get-Location
 	$CmdArray = '7za.exe','dos2unix.exe', 'gawk','jq','plink.exe','pscp.exe'
 
+	$WebClient = New-Object System.Net.WebClient
+	$xmlSecret = [xml](Get-Content $XMLSecretFile)
+	$azureBlobLoc = $xmlSecret.secrets.blobStorageLocation
```
ader1990: change the parameter name to something relevant like: toolsBaseURL (this can be any url), it is not bound to Azure

README.md
```diff
@@ -24,7 +24,7 @@ LISAv2 includes below test suite categories:
 
 3. You must have a valid Windows Azure Subscription.
 
-4. You download 3rd party software in Tools folder.
+4. You download 3rd party software in Tools folder. If you are using Azure blob in Storage account, you can add a tag <blobStorageLocation></blobStorageLocation> in AzureSecret.xml file.
```
ader1990: Why not use "If you have the tools on a webserver accessible by the node where you are running LISAv2, you can set in the secrets the base path of the tools."?

Do not use references to AzureSecret.xml because there is no such hardcoded parameter, it can be a HyperVSecrets file, or AzureStack secrets file, depending on the platform.

RunTests.ps1
ader1990: change the azureBlobLoc var name to a relevant one, as it can be any web base url.

ader1990:
@juhlee-microsoft  From a practical standpoint, even if you remove the tools from the repo using a commit, won't the repo still be subject to licenses from those binary tools? As per git inner workings, anyone can revert to a commit before the removal and find those binaries.

RunTests.ps1
juhlee-microsoft: Fixed in 379457b67f89c8a701fa7d13e4edce9b32372cf2

RunTests.ps1
juhlee-microsoft: This is indirect advertisement of Azure.

README.md
juhlee-microsoft: XML tag is static in the code. but removed the static Secret xml file. Fixed in a1465afd8209c67daeee9c8266be05e3840bfe69

RunTests.ps1
juhlee-microsoft: Again, this is indirect Azure advertisement.

.vscode/settings.json
juhlee-microsoft: Fixed in 991a361aefd32a43fc80a6b056af67fee0715625

RunTests.ps1
ader1990: advertisement for whom? I suggest having a clear and concise code base, not an "advertisement" code base. The name is very confusing and if you want to use only hyper-v testing on a non-azure env, it might be confusing for someone to be left to think that his only option for the tools location is an azure blob.

juhlee-microsoft: My original design was for Azure vs. non-Azure (Hyper-V or local env). Those files should be placed in Tools folder in advance. (README) 

If no files in Tools folder of Hyper-V case, it comes down to 'if' condition's false and jump to exception catch block. (line 518 if condition) 

Just added more clear instruction in 36f2d8dc71deca7bdc30bb54ec36c3e97f3e3e7b

ader1990:
please do a squash on merge

RunTests.ps1
```diff
@@ -93,13 +93,24 @@ try {
...
 			Write-Error "File not found in Tools folder: $_. Testing terminates."
-			throw [System.IO.FileNotFoundException]
+			Write-Host "Downloading required files from Azure blob of your Storage Account"
+
+			$WebClient.DownloadFile("$azureBlobLoc/$_","$CurrentDirectory\Tools\$_")
+
+			if (Test-Path "$CurrentDirectory\Tools\$_") {
+				# Successfully downloaded files
+				Write-Host "File $_ successfully downloaded in Tools folder: $_."				
+			}
```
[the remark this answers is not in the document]
juhlee-microsoft: Fixed - 05083ab5f0e8e8a81702a5e971393df3909e8270

PR #116, opened 2018-07-30
azure: Trim ARMImageName value before usage and added unit simple tests to partially test the affected code

ader1990:
ARMImageName parameter value may come with leading spaces,
which leads to errors in the ARM templates.

This patch makes sure that the ARMImageName value is properly
trimmed before any usage.

### DIFF
--- a/Libraries/Framework.psm1
+++ b/Libraries/Framework.psm1
@@ -32,11 +32,12 @@ Function ValidateParameters()
 		#region Validate Parameters
 		if ( !$ARMImageName -and !$OsVHD )
 		{
-			$ParameterErrors += "-ARMImageName <'Publisher Offer Sku Version'>, or -OsVHD <'VHD_Name.vhd'> is required."
+			$ParameterErrors += "-ARMImageName '<Publisher> <Offer> <Sku> <Version>', or -OsVHD <'VHD_Name.vhd'> is required."
 		}
-		if ($ARMImageName.Split(" ").Count -ne 4)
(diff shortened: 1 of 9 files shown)


Libraries/Framework.psm1
```diff
@@ -32,11 +32,12 @@ Function ValidateParameters()
...
-			$ParameterErrors += "-ARMImageName <'Publisher Offer Sku Version'>, or -OsVHD <'VHD_Name.vhd'> is required."
+			$ParameterErrors += "-ARMImageName '<Publisher> <Offer> <Sku> <Version>', or -OsVHD <'VHD_Name.vhd'> is required."
 		}
-		if ($ARMImageName.Split(" ").Count -ne 4)
+		if ($ARMImageName.Trim().Split(" ").Count -ne 4)
 		{
-			$ParameterErrors += "Invalid value for -ARMImageName <'Publisher Offer Sku Version'> provided. 'Publisher Offer Sku Version' should be separated by space ' ' char."
+			$ParameterErrors += ("Invalid value for the provided ARMImageName parameter: <'${ARMImageName}'>." + `
+                                 "The ARM image should be in the format: '<Publisher> <Offer> <Sku> <Version>'.")
 		}
```
simonxiaoss: This notation ('<Publisher> <Offer> <Sku> <Version>') will be confusing. I assume user need to provide strings without ‘>’ or '<'

juhlee-microsoft: I think it looks ok.

PR #119, opened 2018-07-31
two lines missing in last check-in

juhlee-microsoft:
Saved from last feedback. (The file was not saved in previous check-in)

"|| return" after "cd" command will act successful return to caller if "cd" cmd fails


Testscripts/Linux/run-ntttcp-and-tcping.sh
```diff
@@ -27,15 +27,15 @@
 	if [ "$(which ntttcp)" == "" ]; then
 		rm -rf ntttcp-for-linux
 		git clone https://github.com/Microsoft/ntttcp-for-linux
```
simonxiaoss: we should use the release version, not the code from master branch.

PR #130, opened 2018-08-02
Correcting PR#125

iamshital:
The fix applied in PR#125 is wrong. 
It just hides the problem instead of fixing it. The bug is caused by the empty string in $ARMImageName. 

To resolve this issue, we need to add one more condition to if statement to check if $ARMImageName is empty string.
The new PR is created with correct fix.

PR execution has been verified.

### DIFF
--- a/Libraries/Framework.psm1
+++ b/Libraries/Framework.psm1
@@ -34,7 +34,7 @@ Function ValidateParameters()
 		{
 			$ParameterErrors += "-ARMImageName '<Publisher> <Offer> <Sku> <Version>', or -OsVHD <'VHD_Name.vhd'> is required."
 		}
-		if (($ARMImageName.Trim().Split(" ").Count -ne 4) -and !$OsVHD) 
+		if (($ARMImageName.Trim().Split(" ").Count -ne 4) -and ($ARMImageName -ne "")) 
 		{
 			$ParameterErrors += ("Invalid value for the provided ARMImageName parameter: <'${ARMImageName}'>." + `
                                  "The ARM image should be in the format: '<Publisher> <Offer> <Sku> <Version>'.")


ader1990:
PR#125 is not wrong, it fixes the situation when you have an image string given as a param in this format: `" 
   test test test test     ".

The extra test you are doing is useless: $ARMImageName -ne "", as the first condition will be false in case of empty string ( the count is 1)

Libraries/Framework.psm1
```diff
@@ -34,7 +34,7 @@ Function ValidateParameters()
 		{
 			$ParameterErrors += "-ARMImageName '<Publisher> <Offer> <Sku> <Version>', or -OsVHD <'VHD_Name.vhd'> is required."
 		}
-		if (($ARMImageName.Trim().Split(" ").Count -ne 4) -and !$OsVHD) 
+		if (($ARMImageName.Trim().Split(" ").Count -ne 4) -and ($ARMImageName -ne ""))
```
ader1990: The extra test you are adding is useless: $ARMImageName -ne "", as the first condition will be false in case of empty string ( the count is 1)

iamshital: This PR is in response to the error when $OsVHD value was provided (="Some value") and $ARMImageName="" (empty string).

In this case, ($ARMImageName.Trim().Split(" ").Count -ne 4) case will become true and will generate error - that ARMImageName should be in some specific format. But we do not want this to happen because, we did provided any ARMImageName. It is empty string by default.

We want to to check ($ARMImageName.Trim().Split(" ").Count -ne 4) condition only when the $ARMImageName is not empty.

Now, in $OsVHD ="Some value"and $ARMImageName="" (empty string) scenario,
($ARMImageName.Trim().Split(" ").Count -ne 4) becomes true AND ($ARMImageName -ne "") becomes false. So, this total statement becomes false and we do not see the error message.
In the absence of ($ARMImageName -ne ""), this statement becomes true and we hit the error log and automation fails itself, which we do not want.

ader1990: I understand now, the problem is you want to not show an error when the user leaves the ARMImageName undefined.

ader1990: for a faster execution, put the -ne "" check first

iamshital: This change is to validate parameter - ARMImageName only. It has nothing to do with OsVHD.

juhlee-microsoft: The title should be better. It led the long conversation. This is not correcting the previous PR but another addendum in the logic.

PR #144, opened 2018-08-06
Fixed : LISAv2 not exiting with correct exit code

iamshital:


iamshital:
Execution is verified.

`[INFO ] LISAv2 exits with code: <no exit code>` changed to `[INFO ] LISAv2 exits with code: 0`

RunTests.ps1
```diff
@@ -438,7 +438,7 @@ try {
 		Write-Output "Setting workspace back to original location: $originalWorkingDirectory"
 		Set-Location $originalWorkingDirectory
 	}
-	Get-Variable -Exclude PWD,*Preference | Remove-Variable -Force -ErrorAction SilentlyContinue
+	Get-Variable -Exclude PWD,*Preference,ExitCode | Remove-Variable -Force -ErrorAction SilentlyContinue
```
ader1990: when run manually, this line leaves your environment broken in case you might have had special env variables.
I suggest you to create a inclusive list with all the lisav2 env variables and only remove those.

ader1990: If you do this, it will help "documenting" all of these globals other scripts expect to find.
All in all, using globals is a major lisav2 flaw that needs to be addressed by using function parameters and encapsulation.

iamshital: The original line was - 
`Get-Variable -Scope Global | Remove-Variable -Force -ErrorAction SilentlyContinue`
This keeps the session clean for the next execution. LISAv2 doesn't break when running another iteration in same PS session (and replaces any runtime variables).
But, later, above was changed to -
`Get-Variable -Exclude PWD,*Preference | Remove-Variable -Force -ErrorAction SilentlyContinue`
The removes all the variables from PS session leaving PS session broken for next LISAv2 execution.
Reference: https://github.com/LIS/LISAv2/commit/1a5f69334356fb8a5431f286d1b7cf87d8157b17#diff-59264e449e1f333881e7c9485a2e7a0f

If we want to fix the broken PS session issue, then we can revert it to original line.
The scope of this PR is to fix LISAv2 exit code only.

I agree with documenting on all the Global Variables. But, I also have plan to encapsulate all the global variables in single PS object. We can discuss this over email.

iamshital:
Opening two new issues as per discussion in this thread -
https://github.com/LIS/LISAv2/issues/148
https://github.com/LIS/LISAv2/issues/149

PR #155, opened 2018-08-09
Fixed issue of collecting VM logs

iamshital:
Some distros were failing to execute `dmesg` in the absense of sudo permissions.
Fixed with -RunAsSudo


iamshital:
Fixed issue for distros -
1. SLES12SP3
2. DEBIAN

PR #157, opened 2018-08-13
Fix timeout issue of NESTED-KVM-BASIC-VERIFICATION

lizzha:


XML/TestCases/NestedKvmTests.xml
```diff
@@ -12,6 +12,7 @@
             <param>NestedUserPassword='NESTED_VM_PASSWORD'</param>
             <param>HostFwdPort=60022</param>
         </TestParameters>
+        <Timeout>720</Timeout>
```
juhlee-microsoft: This test directly calls Bash script, and I can not find timeout value in nested_kvm_basic_test.sh. Instead, there is 'retry_times' variable with value 24. Help me understand this test.

iamshital: @juhlee-microsoft 
https://github.com/LIS/LISAv2/blob/master/Libraries/Test-Framework.psm1#L485

iamshital: @lizzha Is this timeout safe for smaller Azure VM sizes also? (1 core / 2 core VMs)

lizzha: yes, the minimum size supported is Standard_E2_v3, I've verified it's OK

juhlee-microsoft: acked

PR #179, opened 2018-08-31
Add Tagging guideline

juhlee-microsoft:
Add the current tags and its format

### DIFF
--- a/Documents/How-to-use.md
+++ b/Documents/How-to-use.md
@@ -134,9 +134,9 @@ Per Category, each XML file has VM name, Resource Group name, etc. We do not rec
             Like 'VERIFY-TEST-SCRIPT-IN-LINUX-GUEST.ps1', the PowerShell script wraps the multiple 
             Bash or Python script as a parameter of 'RunLinuxCmd' function.
     5. Before PR review, we recommend you run script testing in cmdline/API mode. See above instruction.
-    6. Current tags in the Repo: bvt, network, nested, udp, hv_storvsc, linux_image, net, stress, disk, dpdk,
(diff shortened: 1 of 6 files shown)


Documents/How-to-use.md
```diff
@@ -134,6 +134,9 @@ Per Category, each XML file has VM name, Resource Group name, etc. We do not rec
             Like 'VERIFY-TEST-SCRIPT-IN-LINUX-GUEST.ps1', the PowerShell script wraps the multiple 
             Bash or Python script as a parameter of 'RunLinuxCmd' function.
     5. Before PR review, we recommend you run script testing in cmdline/API mode. See above instruction.
+    6. Current tags in the Repo: bvt, network, nested, udp, hv_storvsc, linux_image, net, stress, disk, dpdk,
+        sriov, kvm, btrfs, smb, storage, xfs, boot, pci_hyperv, core, wala, lsvmbus, synthetic, ext4, kvp,
```
chvalean: I don't think file-system types are worth being considered as tags.
Tags would represent test areas or targeted sections - network, storage etc.
Suggestions:
1. remove ext4, xfs, btrfs
2. consolidate "net" into network
3. tcp, udp - are these relevant as tags?
4. linux_image - not relevant.

juhlee-microsoft: Ok, Let's update this document here and create a new PR for actual xml file update.

juhlee-microsoft: Fixed in 5108b56677ca55dc0a15dbf253f87c2da8705767

chvalean:
Approving the changes, we can further consolidate disk into storage and have storage the single label for this, the same we have network.

PR #184, opened 2018-09-04
Corrected function name format suggestion

juhlee-microsoft:
Keep the suggestion for bash and python while we keep the PS in Verb-Entity format.

### DIFF
--- a/Documents/How-to-use.md
+++ b/Documents/How-to-use.md
@@ -148,7 +148,8 @@ Per Category, each XML file has VM name, Resource Group name, etc. We do not rec
         a. Log should be readable to reflect what’s happening in the test execution.
         b. Remove noise. Determine if it would be error or warning.
         c. Exception only in the case of fatal errors.
-    9. Recommended function name format - Function_Name(). Upper letter in each string connected with underscore character.
+    9. Recommended Bash & Python function name format - Function_Name(). Upper letter in each string connected
+        with underscore character. PowerShell function name format remains Verb-Entity() format like Get-VMSize().
 
 ## Support Contact
 


Documents/How-to-use.md
```diff
@@ -148,7 +148,8 @@ Per Category, each XML file has VM name, Resource Group name, etc. We do not rec
         a. Log should be readable to reflect what’s happening in the test execution.
         b. Remove noise. Determine if it would be error or warning.
         c. Exception only in the case of fatal errors.
-    9. Recommended function name format - Function_Name(). Upper letter in each string connected with underscore character.
+    9. Recommended Bash & Python function name format - Function_Name(). Upper letter in each string connected
```
lizzha: we haven't discussed the naming convention for python code, just checked the current python code, PascalCase is used, should we keep as it?

juhlee-microsoft:
No, regarding Python, I’d like to follow Bash script based model.

________________________________

Documents/How-to-use.md
juhlee-microsoft: No, regarding Python, I’d like to follow Bash script based model.

ader1990: The whole story with the code format -- the whole lisav2 code has random formats and styles, ranging from bash written like powershell and viceversa. In my opinion, there is no use of having conventions if there is no way to enforce them. you cannot enforce a convention on more than 100k LOCs if you do not have an automated tool to do that. I would stick to the community conventions for every programming language, like pep8/flake8 for Python, or any other decent style guide for bash and powershell. coming up with new rules on the fly for pythonand powershell alike is just confusing for one who wrote his entire life as pep8 and get-verb enforce.

juhlee-microsoft: The code lines are relatively small still, and we still have a chance to control those function names. Once we set all in consistent format, it would be easier to manage with Community. 

According to pep8, 
"Function names should be lowercase, with words separated by underscores as necessary to improve readability. Variable names follow the same convention as function names.
mixedCase is allowed only in contexts where that's already the prevailing style (e.g. threading.py), to retain backwards compatibility."

Our codes are not following this rule yet. Should we follow pep8 rule for Python scripts? Since it was not strong constraint, I 'would like to keep both Bash and Python in the same manner. @ader1990 ?

PR #185, opened 2018-09-05
Add a new PS script displaying test cases

juhlee-microsoft:
list out test cases
show the counts of test cases per platform
tag information

### DIFF
--- a/Get-LISAv2Statistics.ps1
+++ b/Get-LISAv2Statistics.ps1
@@ -45,7 +45,7 @@ foreach ($fname in $files) {
 
 # Show the statistics information
 Write-Output ""
-Write-Output "===== Test Cases counts per platform ====="
+Write-Output "===== Test Cases Number per platform ====="
 Write-Output ""
 Write-Output "Azure only: $azure_only"
 Write-Output "Hyper-V only: $hyperv_only"
--- a/Get-LISAv2Statistics.ps1
+++ b/Get-LISAv2Statistics.ps1
@@ -1,0 +1,57 @@
+﻿# Linux on Hyper-V and Azure Test Code, ver. 1.0.0
+# Copyright (c) Microsoft Corporation
+
+# Description: This script displays the LISav2 test case statistics and detail tag.
+
+# Read all test case xml files
+$files = Get-ChildItem XML\TestCases\*.xml
+
+$azure_only = 0
+$hyperv_only = 0
+$both_platforms = 0
+
+# Hashtab for tag info collection
+$tags = @{}
+
+Write-Host ""
+“{0,-60} {1,20} {2,15} {3,15} {4,40}” -f` "TestCase", "Platform", "Category", "Area", "Tags"
+Write-Host "----------------------------------------------------------------------------------------------------------------------------------------------------------"
+
+foreach ($fname in $files) {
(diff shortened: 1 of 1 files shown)


Get-LISAv2Statistics.ps1
```diff
@@ -0,0 +1,57 @@
...
+$tags = @{}
+
+Write-Output ""
+“{0,-60} {1,20} {2,15} {3,15} {4,40}” -f` "TestCase", "Platform", "Category", "Area", "Tags"
+Write-Output "----------------------------------------------------------------------------------------------------------------------------------------------------------"
+
+foreach ($fname in $files) {
+    $xml = [xml](Get-Content $fname)
+    foreach ($item in $xml.TestCases.test) {
+        “{0,-60} {1,20} {2,15} {3,15} {4,40}” -f` $item.testName, $item.Platform, $item.Category, $item.Area, $item.Tags
```
ader1990: make sure when you copy code, to replace the non standard quotation marks with standard quotation marks, and not the literature ones.

“ is different than "

also, remove the back tick from -f`

Get-LISAv2Statistics.ps1
```diff
@@ -0,0 +1,57 @@
+﻿# Linux on Hyper-V and Azure Test Code, ver. 1.0.0
+# Copyright (c) Microsoft Corporation
+
+# Description: This script displays the LISav2 test case statistics and detail tag.
```
ader1990: what do you mean by "detail tag"?

Get-LISAv2Statistics.ps1
```diff
@@ -0,0 +1,57 @@
...
+            } else {
+                $tags.Add($single_tag, 1)
+            }
+
+        }
+    }
+}
+
+# Show the statistics information
+Write-Output ""
```
ader1990: Write-Output has a pretty bad way of showing characters, appears as return value from a function and does not show on some console (like jenkins).

Get-LISAv2Statistics.ps1
```diff
@@ -0,0 +1,57 @@
...
+Write-Output "===== Test Cases counts per platform ====="
```
ader1990: Test Cases Number*

Get-LISAv2Statistics.ps1
juhlee-microsoft: Fixed in c583a835aba535a7ccb9ccf1dc847a20e1881fc0

Get-LISAv2Statistics.ps1
juhlee-microsoft: Changed in c583a835aba535a7ccb9ccf1dc847a20e1881fc0

Get-LISAv2Statistics.ps1
juhlee-microsoft: This is a change  based on PSScriptAnalyzer result. No plan to run this script by Jenkins yet.

Get-LISAv2Statistics.ps1
juhlee-microsoft: Fixed in ee06dee7ca4698a2bfa07888ee6c509d081c9b2c

PR #190, opened 2018-09-06
Fixing the issue with RunLinuxCmd

iamshital:
Skip uploading of the runtest.sh file only if the test platform in not HyperV.
In case of HyperV, it must upload the runtest.sh file.

We've started working on AzureStack platform as well, which is similar to Azure. Hence, instead of explicitly checking for Azure, it is better to check for HypeV platform and instruct automation to upload runtest.sh file.

### DIFF
--- a/Libraries/CommonFunctions.psm1
+++ b/Libraries/CommonFunctions.psm1
@@ -1438,7 +1438,8 @@ Function RemoteCopy($uploadTo, $downloadFrom, $downloadTo, $port, $files, $usern
 
 Function WrapperCommandsToFile([string] $username,[string] $password,[string] $ip,[string] $command, [int] $port)
 {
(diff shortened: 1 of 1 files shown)


Libraries/CommonFunctions.psm1
```diff
@@ -1438,7 +1438,8 @@ Function RemoteCopy($uploadTo, $downloadFrom, $downloadTo, $port, $files, $usern
 
 Function WrapperCommandsToFile([string] $username,[string] $password,[string] $ip,[string] $command, [int] $port)
 {
-    if ( ( $lastLinuxCmd -eq $command) -and ($lastIP -eq $ip) -and ($lastPort -eq $port) -and ($lastUser -eq $username) -and ($TestPlatform -eq "Azure"))
+    if ( ( $lastLinuxCmd -eq $command) -and ($lastIP -eq $ip) `-and ($lastPort -eq $port) `
+    -and ($lastUser -eq $username) -and ($TestPlatform -ne "HyperV"))
```
ader1990: leave an indent if you split the line

iamshital: @ader1990 Fixed. Thanks.

PR #203, opened 2018-09-11
Add distro short name and test result term

juhlee-microsoft:
Used case insensitive or irregular pattern of each name. In order to publish our all data in PowerBI or Graph update, we should use the same distro short name.

### DIFF
--- a/Documents/How-to-use.md
+++ b/Documents/How-to-use.md
@@ -153,11 +153,23 @@ Per Category, each XML file has VM name, Resource Group name, etc. We do not rec
         c. Exception only in the case of fatal errors.
     9. Recommended Bash & Python function name format - Function_Name(). Upper letter in each string connected
         with underscore character. PowerShell function name format remains Verb-Entity() format like Get-VMSize().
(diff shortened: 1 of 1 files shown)


Documents/How-to-use.md
```diff
@@ -153,6 +153,12 @@ Per Category, each XML file has VM name, Resource Group name, etc. We do not rec
         c. Exception only in the case of fatal errors.
     9. Recommended Bash & Python function name format - Function_Name(). Upper letter in each string connected
         with underscore character. PowerShell function name format remains Verb-Entity() format like Get-VMSize().
+    10. Use the same terminology: PASS, FAIL, ABORTED (all upper cases)
+
+## Use recommended distro name
+
+    1. Short distro name in log, graph or script: CentOS, ClearLinux, Debian, OpenSUSE, Oracle, RHEL, SLES, Ubuntu, CoreOS.
```
chvalean: Can this be displayed as a list? would make it more visible I would say

juhlee-microsoft: Fixed in 9e8c739427e19176100d4eb850903c9eaf39ec89

PR #213, opened 2018-09-19
TestCase name format should be all in upper cases.

juhlee-microsoft:
TC names should be in all upper cases. Fixed for recent added test cases. 
Verified in Jenkins menu refresh, http://penguinator.westus2.cloudapp.azure.com/view/Microsoft/job/Microsoft-Refresh-Test-Selection-Menus/40/console


XML/TestCases/FunctionalTests.xml
```diff
@@ -59,7 +59,7 @@
         <Tags>kvp</Tags>
     </test>
     <test>
-        <testName>KVP-ModifyKeyValues</testName>
+        <testName>KVP-MODIFYKEYVALUES</testName>
```
lizzha: suggest to add dashes to separate words, KVP-MODIFY-KEY-VALUES, which makes it more readable

lizzha: apply to other cases

juhlee-microsoft: Fixed in c27ae1813b48017fdf77bccd9f2689caf43aaba4 & verified in http://penguinator.westus2.cloudapp.azure.com/view/Microsoft/job/Microsoft-Refresh-Test-Selection-Menus/42/console

juhlee-microsoft: Added three people who made them in history. The notification will send out.

PR #223, opened 2018-09-21
kdump: Fixed test case names

mbivolan:


juhlee-microsoft:
@mbivolan It is supposed to use '-' dash not '_' underline.

mbivolan:
@juhlee-microsoft  Replaced all '_' with '-'

PR #237, opened 2018-09-28
Notify in case specified test not found

santoshx:
This will be useful in case when we mistype a test name.


RunTests.ps1
```diff
@@ -197,6 +197,10 @@ try {
 	#Validate all XML files in working directory.
 	$allTests = CollectTestCases -TestXMLs $TestXMLs
 
+	if( !$allTests.innerXML ) {
+		LogErr "Specified test not found!"
```
juhlee-microsoft: Improve log contents for better understanding of error case.
Also add throw statement with exit code, so that it will stop running.

PR #241, opened 2018-10-01
hyperv: Fix Hyper-V group cleanup

ader1990:


Libraries/HyperV.psm1
```diff
@@ -239,86 +239,41 @@ Function CreateAllHyperVGroupDeployments($setupType, $xmlConfig, $Distro, $Debug
...
-{
-    try
-    {
-        $AllGroups = $null
-        LogMsg "Checking if HyperV VM group '$HyperVGroupName' exists in $HyperVHost..."
-        $AllGroups = Get-VMGroup -Name $HyperVGroupName -ErrorAction SilentlyContinue -ComputerName $HyperVHost
+Function DeleteHyperVGroup([string]$HyperVGroupName, [string]$HyperVHost) {
+    if ($ExistingRG) {
+        LogMsg "Skipping removal of Hyper-V VM group ${HyperVGroupName}"
+        return $true
```
juhlee-microsoft: Does it make sense with $true? This function removes Hyper-V Group if exists. I suggest skipping could be '$false'. Thought?

ader1990: the method should return false on failure only, it is mandatory to return true there, as there is no failure

PR #245, opened 2018-10-03
enableRoot: Do sync to fix #244

### LINKED ISSUE #244: HyperV guest no longer ssh reachable after applying just created ICABase checkpoint
This is happening because we just did operation on /etc/ssh/sshd_config and then immediately created a checkpoint with -Force option. This is resulting in corrupted or no /etc/ssh/sshd_config on guest.

santoshx:
This fixes an issue where the ssh daemon config file could remain blank after the changes.
Adding a sync command to flush all the changes to disk.


chvalean:
Thank you @santoshx for the fix.
As this is an important fix, I've added a description of the issue, I would recommend for you to do the same for future code for better understanding of the changes.

PR #238, opened 2018-09-28
Hyper-V: Added 14 NET test cases and support for secondary VM

adriansuhov:
This adds the following test cases:
--NET-FILE-COPY-SCP
--NET-IP-INJECTION
--NET-TCP-CORRUPTION
--NET-PRIVATE
--NET-CHANGE-PRIVATE-NIC
--NET-JUMBO-FRAMES
--NET-COPY-LARGE-FILE
--NET-COPY-FILE-DIFFERENT-MTU
--NET-COPY-BOTH-WAYS
--NET-VLAN-TAGGING
--NET-VLAN-TRUNKING
--NET-VXLAN
--NET-MULTICAST
--ETHTOOL-CHECK-STATISTICS

This also introduces a change in the framework that will give the
possibility to run a dependency VM when needed. The framework can now
handle the switch from a setup type to another. Before this, on Hyper-V,
all the tests were performed on the setup type of the first test in
queue.

::**TEST RESULTS (all NET area) on Xenial**::
Total Execution Time(dd:hh:mm) 0:2:1
XML file: TestConfiguration

#Sr. Test Name : Test Result : Test Duration (in minutes)
----------------------------------------------------
1. NET-EXTERNAL : PASS : 4.85
2. NET-INTERNAL : PASS : 2.7
3. NET-LEGACY : PASS : 2.94
4. NET-PROMISCUOUS : PASS : 2.36
5. NET-OPERSTATE : PASS : 2.03
6. NET-CHANGE-INTERNAL-NIC : PASS : 2.69
7. NET-MAX-SYNTHETIC-NICS : PASS : 3.84
8. NET-MAX-LEGACY-NICS : PASS : 3.15
9. NET-MAX-NICS : PASS : 5.57
10. NET-FILE-COPY-SCP : PASS : 10.51
11. NET-IP-INJECTION : PASS : 1.31
12. NET-TCP-CORRUPTION : PASS : 3.02
13. NET-PRIVATE : PASS : 6.81
14. NET-CHANGE-PRIVATE-NIC : PASS : 4.26
15. NET-JUMBO-FRAMES : PASS : 6.31
16. NET-COPY-LARGE-FILE : PASS : 4.78
17. NET-COPY-FILE-DIFFERENT-MTU : PASS : 4.75
18. NET-COPY-BOTH-WAYS : PASS : 23.76
19. NET-VLAN-TAGGING : PASS : 5.56
20. NET-VLAN-TRUNKING : PASS : 5.37
[...]

PR #246, opened 2018-10-03
Fixed a breaking issue introduced in PR #238

iamshital:
Variable VmSetup is only defined for HyperV but it is used irrespective of Test Platform which is causing referring to null array when running Azure Test cases.


chvalean:
Merging this due to impact, will have to further double-check the automation status

PR #254, opened 2018-10-09
CommonFunctions.psm1: Rename function Run-Test to Run-OnLinux

santoshx:
This function is actually to run bash script on Linux. 
There is another function Run-Test in Test-Framework.psm1 to run test.
Grep through the entire code did not find anywhere it is currently in use. Hence safe to rename.


iamshital:
> Grep through the entire code did not find anywhere it is currently in use. Hence safe to rename.

@santoshx If it not used anywhere, we should **remove** it.

iamshital:
@juhlee-microsoft @lizzha @chvalean I propose to remove this entire function.

santoshx:
This function is very useful. I'm able to save lot of typing & duplication. We should keep it. I think it was not in use because of duplicate name.

chvalean:
Hi @santoshx 
You should try to use function RunLinuxCmd in your tests, as that is the reference one for such purposes.
Are you using function Run-Test in any of your tests forked, because we don't see it in use in lisav2.

chvalean:
> @chvalean
> Seems like RunLinuxCmd is to run _a_ command.
> This Run-Test (Run-OnLinux) is to run a script.
> I'm going to use it.
> If removed I'll define locally in my .ps1, your call.

RunLinuxCmd can be well used to trigger a script located inside the VM, please refer to the below sample script that shows that.
I was not trying to push the removal of Run-Test (Run-OnLinux), however as RunLinuxCmd is more versatile and has various parameters, it would keep the code more consistent in how commands or scripts are ran from the host on a gust VM.

https://github.com/LIS/LISAv2/blob/master/Testscripts/Windows/VERIFY-NVIDIA-CUDA-DRIVER-INSTALLATION.ps1#L27

juhlee-microsoft:
I recommend to use the existing function if you can. Renaming is risky.

santoshx:
> > @chvalean
> > Seems like RunLinuxCmd is to run _a_ command.
> > This Run-Test (Run-OnLinux) is to run a script.
> > I'm going to use it.
> > If removed I'll define locally in my .ps1, your call.
> 
> RunLinuxCmd can be well used to trigger a script located inside the VM, please refer to the below sample script that shows that.
> I was not trying to push the removal of Run-Test (Run-OnLinux), however as RunLinuxCmd is more versatile and has various parameters, it would keep the code more consistent in how commands or scripts are ran from the host on a gust VM.
> 
> https://github.com/LIS/LISAv2/blob/master/Testscripts/Windows/VERIFY-NVIDIA-CUDA-DRIVER-INSTALLATION.ps1#L27

Yes, I see that RunLinuxCmd can be used to run the script on guest but we need to call it several times to achieve.
Instead we can call function Run-BashScript and it takes care of how to run it on guest.

santoshx:
> I recommend to use the existing function if you can. Renaming is risky.

There is same function name 'Run-Test' in two files CommonFunctions.psm1 & Test-Framework.psm1.

santoshx:
Close this for now. I'll come up with better fix. Thank you for your comments.

PR #255, opened 2018-10-09
hyper-v: Cache vhds if the source path is UNC

ader1990:
If the OSVhd path is a local file, resume to the current workflow.

If the OSVhd path is an UNC path (ex. \\smb_server\smb_share), copy the VHD to the Hyper-V vhdx store and use that path for the parent vhd (if the VHD is not there already).

The unicity of the VHD will be computed according to the VHD disk identifier which will be appended to the VHD cached path.

### DIFF
--- a/Libraries/HyperV.psm1
+++ b/Libraries/HyperV.psm1
@@ -337,9 +337,35 @@ Function CreateHyperVGroupDeployment([string]$HyperVGroup, $HyperVGroupNameXML, 
                 $CurrentVMOsVHDPath = "$DestinationOsVHDPath\$HyperVGroupName-role-$i-diff-OSDisk${vhdSuffix}"
                 $i += 1
             }
-            $Out = New-VHD -ParentPath "$SourceOsVHDPath\$OsVHD" -Path $CurrentVMOsVHDPath -ComputerName $HyperVHost
-            if ($?)
-            {
+
+            $parentOsVHDPath = $OsVHD
+            if ($SourceOsVHDPath) {
+                $parentOsVHDPath = Join-Path $SourceOsVHDPath $OsVHD
+            }
+            $infoParentOsVHD = Get-VHD $parentOsVHDPath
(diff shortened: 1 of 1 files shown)


juhlee-microsoft:
Last time, @iamshital set up a FTP server. Was it for file transfer?

Libraries/HyperV.psm1
```diff
@@ -337,9 +337,36 @@ Function CreateHyperVGroupDeployment([string]$HyperVGroup, $HyperVGroupNameXML,
...
+                LogMsg "Checking if we have a local VHD with the same disk identifier on the host"
+                $hypervVHDLocalPath = (Get-VMHost -ComputerName $HyperVHost).VirtualHardDiskPath
+                $vhdName = [System.IO.Path]::GetFileNameWithoutExtension($(Split-Path -Leaf $parentOsVHDPath))
+                $newVhdName = "{0}-{1}{2}" -f @($vhdName, $infoParentOsVHD.DiskIdentifier.Replace("-", ""),$vhdSuffix)
+                $localVHDPath = Join-Path $hypervVHDLocalPath $newVhdName
+                if ((Test-Path $localVHDPath)) {
+                    LogMsg "${parentOsVHDPath} is already found at path ${localVHDPath}"
+                } else {
+                    LogMsg "${parentOsVHDPath} will be copied at path ${localVHDPath}"
+                    Copy-Item -Force $parentOsVHDPath $localVHDPath
```
juhlee-microsoft: This Copy-Item runs over remote connection and can be hug size.  Should we care with md5sum?

chvalean: this is the actual use case - copy vhd files from the internal share.
This has nothing to do with FTP repo, that would be something else.

Adrian is actually checking the DiskIdentifier of the VHD, and that is unique. MD5 was our initial thought as well, but then we went with DiskIdentifier as it is a feature of VHD.

PR #256, opened 2018-10-09
framework: Readded nc.exe as a dependency

adriansuhov:
Readding nc.exe as a dependency. Also updated readme file with the 3d
party software needed in Tools folder


juhlee-microsoft:
Change looks ok, but @adriansuhov can you post the screenshot of your Tools folder after a simple test executed in your working branch. Running test from master or other branches may not show nc.exe. Your working branch is the only place shows this binary.

PR #258, opened 2018-10-09
Usage fix

iamshital:


juhlee-microsoft:
It looks ok, but can you post the test result?

iamshital:
The code is already is being used on daily basis to send out emails to LISDEV team.
And you can see current results here : Utilities/job/tool-monitor-subscription-usage/

PR #261, opened 2018-10-10
Resolve issue for running code remotely

LiliDeng:
Resolve issue for running code remotely when running machine's switch inconsistent with remote ones.


LiliDeng:
@juhlee-microsoft azure cases don't go here

PR #272, opened 2018-10-11
Add keywords for COREOS when check dmesg to ignore them.

v-sijliu:
Confirmed with Distro vendor.


XML/Other/ignorable-boot-errors.xml
```diff
@@ -26,5 +27,8 @@
             <keywords>Broken pipe</keywords>
             <keywords>errors=remount-ro</keywords>			
         </errors>
-        <warnings/>
+        <warnings>
+            <keywords>Proceeding WITHOUT firewalling in effect!</keywords>
+            <keywords>7 urandom warning(s) missed due to ratelimiting</keywords>
```
lizzha: better to set to "urandom warning(s) missed due to ratelimiting" to be more generic

PR #278, opened 2018-10-12
Corrected the tag in nested tests

juhlee-microsoft:


XML/TestCases/NestedVmTests.xml
```diff
@@ -210,7 +210,7 @@
         <Platform>HyperV</Platform>
         <Category>Performance</Category>
         <Area>Nested</Area>
-        <Tags>nested,hyperv,win</Tags>
+        <Tags>nested,kvm</Tags>
```
lizzha: It's nested Hyper-V case, should not use the tag of kvm. Since we're still discussing the tag in email threads, I suggest to close this PR, and make the change after it's finalized

PR #282, opened 2018-10-15
kernel_install: Added support for multiple kernel install files

ader1990:
This patch allows multiple kernel files to be installed -- for example hyperv daemons, image or debug kernel.

Example Run-Lisav2 argument:
-CustomKernel "localfile:C:\debfolders\*.deb". Now lisav2 will make sure it uploads all the files and installs them.
Works with -CustomKernel "localfile:C:\debfolders\*.rpm" too.

### DIFF
--- a/Libraries/CommonFunctions.psm1
+++ b/Libraries/CommonFunctions.psm1
@@ -319,20 +319,30 @@ function InstallCustomKernel ($CustomKernel, $allVMData, [switch]$RestartAfterUp
 			$jobCount = 0
 			$kernelSuccess = 0
 			$packageInstallJobs = @()
+			$CustomKernelLabel = $CustomKernel
 			foreach ( $vmData in $allVMData )
 			{
 				RemoteCopy -uploadTo $vmData.PublicIP -port $vmData.SSHPort -files ".\Testscripts\Linux\$scriptName,.\Testscripts\Linux\DetectLinuxDistro.sh" -username $user -password $password -upload
-				if ( $CustomKernel.StartsWith("localfile:"))
-				{
+				if ( $CustomKernel.StartsWith("localfile:")) {
 					$customKernelFilePath = $CustomKernel.Replace('localfile:','')
-					RemoteCopy -uploadTo $vmData.PublicIP -port $vmData.SSHPort -files ".\$customKernelFilePath" -username $user -password $password -upload
+					$customKernelFilePath = (Resolve-Path $customKernelFilePath).Path
+					if ($customKernelFilePath -and (Test-Path $customKernelFilePath)) {
+						RemoteCopy -uploadTo $vmData.PublicIP -port $vmData.SSHPort -files $customKernelFilePath `
+							-username $user -password $password -upload
+					} else {
+						LogErr "Failed to find kernel file ${customKernelFilePath}"
+						return $false
+					}
+					$CustomKernelLabel = "localfile:{0}" -f @((Split-Path -Leaf $CustomKernel.Replace('localfile:','')))
 				}
 				RemoteCopy -uploadTo $vmData.PublicIP -port $vmData.SSHPort -files ".\Testscripts\Linux\$scriptName,.\Testscripts\Linux\DetectLinuxDistro.sh" -username $user -password $password -upload
 
 				$Null = RunLinuxCmd -ip $vmData.PublicIP -port $vmData.SSHPort -username $user -password $password -command "chmod +x *.sh" -runAsSudo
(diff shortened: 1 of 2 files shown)


Libraries/CommonFunctions.psm1
```diff
@@ -319,20 +319,31 @@ function InstallCustomKernel ($CustomKernel, $allVMData, [switch]$RestartAfterUp
...
 			$packageInstallJobs = @()
+			$CustomKernelLabel = $CustomKernel
 			foreach ( $vmData in $allVMData )
 			{
 				RemoteCopy -uploadTo $vmData.PublicIP -port $vmData.SSHPort -files ".\Testscripts\Linux\$scriptName,.\Testscripts\Linux\DetectLinuxDistro.sh" -username $user -password $password -upload
 				if ( $CustomKernel.StartsWith("localfile:"))
 				{
 					$customKernelFilePath = $CustomKernel.Replace('localfile:','')
-					RemoteCopy -uploadTo $vmData.PublicIP -port $vmData.SSHPort -files ".\$customKernelFilePath" -username $user -password $password -upload
+                    $customKernelFilePath = (Resolve-Path $customKernelFilePath).Path
```
iamshital: Fix the indents.

ader1990:
@iamshital  Execution verified on Ubuntu 16.04 and CentOS 7.5:

Libraries/CommonFunctions.psm1
ader1990: this is how it looks when using tabs, it s a github problem. I warned you when you decided switching to tabs is a good idea that this might happen.

lizzha: looks like the indents are spaces, not tabs

iamshital: > this is how it looks when using tabs, it s a github problem. I warned you when you decided switching to tabs is a good idea that this might happen.

@ader1990 Are you sure that I decided to change the to tabs?

iamshital:
@ader1990 
>this is how it looks when using tabs, it s a github problem. I warned you when you decided switching to tabs is a good idea that this might happen.

What makes you think that I made this change of space to tabs?

iamshital:
@ader1990 As this code touches affects lot of distros, I request you to submit the verification results for other supported distros as well, wherever applicable.

Testscripts/Linux/customKernelInstall.sh
```diff
@@ -180,19 +196,25 @@ InstallKernel()
...
+                    LogMsg "Installing ${CustomKernel##*/}"
+                    rpm -ivh "${CustomKernel##*/}"  >> $logFolder/build-CustomKernel.txt 2>&1
+                    kernelInstallStatus=$?
                 else
-                        prefix="localfile:"
-                        LogMsg "Installing ${CustomKernel#$prefix}"
-                        rpm -ivh "${CustomKernel#$prefix}"  >> $logFolder/build-CustomKernel.txt 2>&1
-                        kernelInstallStatus=$?
+                    LogMsg "Removing packages that do not allow the kernel to be installed"
+                    yum remove -y hypervvssd hypervkvpd hypervfcopyd hyperv-daemons
```
iamshital: Is this CentOS7.5 specific or general requirement to remove these packages before installing kernel?

iamshital: Also, what if, user just provides only one rpm file (just the kernel), in that case removing these packages will be unnecessary.
Please handle this scenario as well.

Libraries/CommonFunctions.psm1
ader1990: @iamshital I fixed the indents, my bad, I thought they were already tabs

Testscripts/Linux/customKernelInstall.sh
ader1990: will be done

ader1990:
@iamshital the code has been tested with ubuntu xenial and centos 7 (which is the use case where we need the change). what were the distros where you require to have the localfile option working? I do not see any documentation on which distros the customKernelInstall should work or pipelines that already use this feature.

chvalean:
@iamshital please approve the PR changes if there are no further comments.
Looking to get this merged so that we can proceed with the rest of the planning.
Will be adding other corner scenarios to be supported when time comes.

iamshital:
> @iamshital please approve the PR changes if there are no further comments.
> Looking to get this merged so that we can proceed with the rest of the planning.
> Will be adding other corner scenarios to be supported when time comes.

@chvalean Understood.

PR #287, opened 2018-10-17
hyper-v framework: deploy 2nd VM to a different host

adriansuhov:
This adds a new option to deploy a VM on a secondary host. How this
works:
-- On a test configuration, if DeployOnDifferentHyperVHost tag is
present in a VMSetup and is set to "yes", the VM will be deployed on the secondary host
specified in GlobalConfigurations.xml.
-- At runtime, the -TestLocation param has to be "host_1,host_2" for
this to be triggered.
Azure code not changed in this PR. Also, all the existing run methods on
Hyper-V are not affected

Also added 1 SR-IOV TC to validate this framework change.

### Test results (2 VMs on different hosts): 
ICA test run on 10/17/2018 07:28:01
VHD under test **Centos_7.5_x64.vhd**
ARM Image under test  :  :  :
Total Executed TestCases        1 (       1 Pass,        0 Fail,        0 Abort)
Total Execution Time(dd:hh:mm) 0:0:10
XML file: TestConfiguration

#Sr. Test Name : Test Result : Test Duration (in minutes)
----------------------------------------------------
1. SRIOV-DISABLEVF-PING : PASS : 10.42

VHD under test **ubuntu_18.04_x64.vhd**
ARM Image under test  :  :  :
Total Executed TestCases        1 (       1 Pass,        0 Fail,        0 Abort)
Total Execution Time(dd:hh:mm) 0:0:12
XML file: TestConfiguration

#Sr. Test Name : Test Result : Test Duration (in minutes)
----------------------------------------------------
1. SRIOV-DISABLEVF-PING : PASS : 12.42

### Test results on existing TC, without DeployOnDifferentHyperVHost tag (2 VMs on same host)
VHD under test Centos_7.5_x64.vhd
ARM Image under test  :  :  :
Total Executed TestCases        1 (       1 Pass,        0 Fail,        0 Abort)
Total Execution Time(dd:hh:mm) 0:0:7
XML file: TestConfiguration

#Sr. Test Name : Test Result : Test Duration (in minutes)
----------------------------------------------------
1. NET-PRIVATE : PASS : 7.47


LiliDeng:
@adriansuhov Hi, your code can deploy 2 VMs under one group against 2 hosts? Am I correct? Current code can support deploy 2 VMs under 2 groups against 2 hosts, can current code meet your requirement?

lizzha:
Deploying two VMs on two Hyper-V server is already supported, can you explain more about the difference in this change?

adriansuhov:
LiliDeng - you are correct on the deployment. It's deploying the 2nd VM inside a group on a different host.
The current automation can't cover the following use case: 
A test that needs to run both on Azure and Hyper-V and needs 2 VMs (on separate hosts for Hyper-V). Right now we have SetupTypes that cover 2 or more VMs. And yes, if they're on different HyperV Groups they will be deployed on a different server. But in this case, on Azure we'll have different Resource Groups. So we'll have 2 VMs both on Azure and Hyper-V, but the VMs on Azure couldn't be used for network tests (they have to be on the same resource group). Because of this behavior, I was forced to cover the case where I wanted the VMs to remain in the same Azure Resource Group but also to be deployed on a different Hyper-V Host if needed by the user.

PR #289, opened 2018-10-17
utils.sh: Set LIS_HOME to location where files copied to

santoshx:
I'm running the tests with non root user ( a sudo user). 
In this case I see that test files are copied to /home/$username not to /root
However when scripts are run on guest as sudo user the $HOME will be /root and files are not there in /root

### DIFF
--- a/Testscripts/Linux/utils.sh
+++ b/Testscripts/Linux/utils.sh
@@ -27,7 +27,7 @@ IFS=$' \t\n'
 # All vars are first defined here
 
 # Directory containing all files pushed by LIS framework
-declare LIS_HOME="$HOME"
+declare LIS_HOME=$(pwd)
 
 # LIS state file used by powershell to get the test's state
 declare __LIS_STATE_FILE="$LIS_HOME/state.txt"


lizzha:
the change might affect many test cases running on Hyper-V, @santoshx have you run regression test against all the affected cases?

santoshx:
> the change might affect many test cases running on Hyper-V, @santoshx have you run regression test against all the affected cases?

I've not run regression. Currently I'm unable to run regression due to by infra issues.

juhlee-microsoft:
> > the change might affect many test cases running on Hyper-V, @santoshx have you run regression test against all the affected cases?
> 
> I've not run regression. Currently I'm unable to run regression due to my infra issues.

I am not sure if we can prove this change is valid. How soon can you post the result?

santoshx:
> > > the change might affect many test cases running on Hyper-V, @santoshx have you run regression test against all the affected cases?
> > 
> > 
> > I've not run regression. Currently I'm unable to run regression due to my infra issues.
> 
> I am not sure if we can prove this change is valid. How soon can you post the result?

@juhlee-microsoft  
I feel change is needed, but I may be wrong, as I'm new to this Framework.
Here what I see, Framework copies files to home directory of user. 
But when a command is run as sudo, the $HOME is /root .
I want to get this through regression.
What is the minimum PC configuration is required for a regression?

juhlee-microsoft:
> > > > the change might affect many test cases running on Hyper-V, @santoshx have you run regression test against all the affected cases?
> > > 
> > > 
> > > I've not run regression. Currently I'm unable to run regression due to my infra issues.
> > 
> > 
> > I am not sure if we can prove this change is valid. How soon can you post the result?
> 
> @juhlee-microsoft
> I feel change is needed, but I may be wrong, as I'm new to this Framework.
> Here what I see, Framework copies files to home directory of user.
> But when a command is run with sudo, the $HOME is /root .
> I want to get this through regression.
> What is the minimum PC configuration required for a regression?

@iamshital @rmsivakanth can you assist to @santoshx ?

chvalean:
@santoshx the behavior you describe and the fix for it look good.
If you can only please run several tests that make use of scripts that get executed as sudo, then that would give us enough validation to make sure that the changes are not breaking anything.
We're also running the functional tests for this PR.

Thank you!

santoshx:
> @santoshx the behavior you describe and the fix for it look good.
> If you can only please run several tests that make use of scripts that get executed as sudo, then that would give us enough validation to make sure that the changes are not breaking anything.
> We're also running the functional tests for this PR.
> 
> Thank you!

@chvalean 
We'll do some relevant tests, next week.

chvalean:
@santoshx  please rebase the PR so we can merge it.
Changes look fine, we ran the functional tests and see no test failures to point out any issues.

Internal reference: "job/PR-303/1271/console"

santoshx:
> @santoshx please rebase the PR so we can merge it.
@chvalean  Done

PR #290, opened 2018-10-17
Replaced with ThrowException fn

juhlee-microsoft:
Using the built-in function ThrowException $_ instead of missing or various format.


lizzha:
Looks OK, do you have the test log when an exception is triggered?

juhlee-microsoft:
> Looks OK, do you have the test log when an exception is triggered?

No, can you make a suggestion? I don't have any top of my head

PR #298, opened 2018-10-18
Revert "Replaced with ThrowException fn (#290)"

iamshital:
This reverts commit 703394e650a4360612416d93c9fc791e6bf03dd5.


juhlee-microsoft:
This does not look right. Simple revert is not a right solution. You will need to find the exact broken part. @iamshital

PR #306, opened 2018-10-23
Migrate Get-SubscriptionUsageTopVMs from JenkinsPipeline and unify AzureSecretsFile usage in Utilities folder.

seansp:


Utilities/Get-AzureVMs.ps1
```diff
@@ -13,35 +12,32 @@ param
 )
 
 #Load libraries
-Get-ChildItem .\Libraries -Recurse | Where-Object { $_.FullName.EndsWith(".psm1") } | ForEach-Object { Import-Module $_.FullName -Force -Global }
+Get-ChildItem ..\Libraries -Recurse | Where-Object { $_.FullName.EndsWith(".psm1") } | ForEach-Object { Import-Module $_.FullName -Force -Global }
 
-if( $NoSecrets -eq $false )
+#Read secrets file and terminate if not present.
```
lizzha: why removing $NoSecrets? it will be not supported to run the tool on local environment without secret file

PR #320, opened 2018-10-26
Support performance testing against CoreOS

LiliDeng:
Run below cases against CoreOS CoreOS Stable Latest and canonical ubuntuserver 18.04-lts Latest
PERF-NETWORK-TCP-NETPERF-SINGLE-PPS-Synthetic
PERF-NETWORK-TCP-NETPERF-MAX-PPS-Synthetic
PERF-NETWORK-TCP-SINGLE-CONNECTION-THROUGHPUT-Synthetic
PERF-STORAGE-4K-IO
PERF-NETWORK-TCP-LATENCY-MULTICONNECTION-Synthetic
PERF-NETWORK-TCP-THROUGHPUT-MULTICONNECTION-NTTTCP-Synthetic

### DIFF
--- a/Testscripts/Linux/perf_fio.sh
+++ b/Testscripts/Linux/perf_fio.sh
@@ -97,12 +97,12 @@ RunFIO()
 	io_increment=128
 
 	####################################
-	echo "Test log created at: ${LOGFILE}"
-	echo "===================================== Starting Run $(date +"%x %r %Z") ================================"
-	echo "===================================== Starting Run $(date +"%x %r %Z") script generated 2/9/2015 4:24:44 PM ================================" >> $LOGFILE
+	LogMsg "Test log created at: ${LOGFILE}"
+	LogMsg "===================================== Starting Run $(date +"%x %r %Z") ================================"
+	LogMsg "===================================== Starting Run $(date +"%x %r %Z") script generated 2/9/2015 4:24:44 PM ================================"
 
 	chmod 666 $LOGFILE
-	echo "--- Kernel Version Information ---" >> $LOGFILE
+	LogMsg "--- Kernel Version Information ---"
 	uname -a >> $LOGFILE
 	cat /proc/version >> $LOGFILE
 	if [ -f /usr/share/clear/version ]; then
@@ -110,7 +110,7 @@ RunFIO()
 	elif [ -f /etc/*-release ]; then
 		cat /etc/*-release >> $LOGFILE
 	fi
-	echo "--- PCI Bus Information ---" >> $LOGFILE
+	LogMsg "--- PCI Bus Information ---"
 	lspci >> $LOGFILE
 	df -h >> $LOGFILE
 	fio --cpuclock-test >> $LOGFILE
@@ -137,10 +137,10 @@ RunFIO()
 				#mkdir $blk_operation
 				#blktrace -w 40 -d /dev/sdf -D $blk_operation &
 				#blktrace -w 40 -d /dev/sdm -D $blk_operation &
-				echo "-- iteration ${iteration} ----------------------------- ${testmode} test, ${io}K bs, ${Thread} threads, ${numjobs} jobs, 5 minutes ------------------ $(date +"%x %r %Z") ---" >> $LOGFILE
+				LogMsg "-- iteration ${iteration} ----------------------------- ${testmode} test, ${io}K bs, ${Thread} threads, ${numjobs} jobs, 5 minutes ------------------ $(date +"%x %r %Z") ---"
 				LogMsg "Running ${testmode} test, ${io}K bs, ${Thread} threads ..."
 				jsonfilename="${JSONFILELOG}/fio-result-${testmode}-${io}K-${Thread}td.json"
-				echo "${fio_cmd} $FILEIO --readwrite=${testmode} --bs=${io}K --runtime=${ioruntime} --iodepth=${Thread} --numjobs=${numjobs} --output-format=json --output=${jsonfilename} --name='iteration'${iteration}" >> $LOGFILE
+				LogMsg "${fio_cmd} $FILEIO --readwrite=${testmode} --bs=${io}K --runtime=${ioruntime} --iodepth=${Thread} --numjobs=${numjobs} --output-format=json --output=${jsonfilename} --name='iteration'${iteration}"
 				$fio_cmd $FILEIO --readwrite=$testmode --bs=${io}K --runtime=$ioruntime --iodepth=$Thread --numjobs=$numjobs --output-format=json --output=$jsonfilename --name="iteration"${iteration} >> $LOGFILE
 				#fio $FILEIO --readwrite=$testmode --bs=${io}K --runtime=$ioruntime --iodepth=$Thread --numjobs=$numjobs --name="iteration"${iteration} --group_reporting >> $LOGFILE
 				iostatPID=`ps -ef | awk '/iostat/ && !/awk/ { print $2 }'`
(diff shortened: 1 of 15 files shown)


Testscripts/Linux/perf_fio.sh
```diff
@@ -102,11 +102,6 @@ RunFIO()
 	echo "===================================== Starting Run $(date +"%x %r %Z") script generated 2/9/2015 4:24:44 PM ================================" >> $LOGFILE
 
 	chmod 666 $LOGFILE
-	echo "Preparing Files: $FILEIO"
-	echo "Preparing Files: $FILEIO" >> $LOGFILE
-	LogMsg "Preparing Files: $FILEIO"
-	# Remove any old files from prior runs (to be safe), then prepare a set of new files.
-	rm fiodata
```
juhlee-microsoft: any reason deleting this line?

Testscripts/Linux/perf_fio.sh
```diff
@@ -153,19 +140,22 @@ RunFIO()
 				echo "-- iteration ${iteration} ----------------------------- ${testmode} test, ${io}K bs, ${Thread} threads, ${numjobs} jobs, 5 minutes ------------------ $(date +"%x %r %Z") ---" >> $LOGFILE
 				LogMsg "Running ${testmode} test, ${io}K bs, ${Thread} threads ..."
 				jsonfilename="${JSONFILELOG}/fio-result-${testmode}-${io}K-${Thread}td.json"
-				fio $FILEIO --readwrite=$testmode --bs=${io}K --runtime=$ioruntime --iodepth=$Thread --numjobs=$numjobs --output-format=json --output=$jsonfilename --name="iteration"${iteration} >> $LOGFILE
+				echo "${fio_cmd} $FILEIO --readwrite=${testmode} --bs=${io}K --runtime=${ioruntime} --iodepth=${Thread} --numjobs=${numjobs} --output-format=json --output=${jsonfilename} --name='iteration'${iteration}" >> $LOGFILE
```
juhlee-microsoft: @LiliDeng There are many 'echo' statement in this script. If needed, can we placed with LogMsg? It intends to keep the log for later review. 'echo' output will be gone after run-time.

Testscripts/Linux/perf_fio.sh
```diff
@@ -203,20 +193,6 @@ CreateRAID0()
 	LogMsg "INFO: Creating RAID of ${count} devices."
 	sleep 1
 	mdadm --create ${mdVolume} --level 0 --raid-devices ${count} /dev/sd[c-z][1-5]
-	sleep 1
```
juhlee-microsoft: Help me understand why those lines removal

Testscripts/Linux/perf_fio.sh
```diff
@@ -245,16 +221,6 @@ CreateLVM()
 	pvcreate /dev/sd[c-z][1-5]
 	vgcreate ${vggroup} /dev/sd[c-z][1-5]
 	lvcreate -l 100%FREE -i 12 -I 64 ${vggroup} -n lv1
-	time mkfs -t $1 -F /dev/${vggroup}/lv1
```
juhlee-microsoft: ditto

Testscripts/Linux/perf_netperf.sh
```diff
@@ -92,56 +91,63 @@ fi
...
     max_port=30031
     while [ $current_port -le $max_port ]; do
-        server_command="netserver -p $current_port > netperf-server-output.txt 2>&1"
+        server_command="nohup ${netperf_cmd} netserver -p $current_port -D > netperf-server-output.txt 2>&1 &"
         ssh ${server} $server_command
         LogMsg "Executed: $server_command"
         current_port=$(($current_port+1))
     done
-    #Start the sar monitor on server.
-    server_sar_command="sar -n DEV 1 ${test_duration} > netperf-server-sar-output.txt"
```
juhlee-microsoft: This monitoring statement is needed.

Testscripts/Windows/PERF-STORAGE-MULTIDISK-RAID0-FIO.ps1
```diff
@@ -46,7 +46,7 @@ chmod 666 /root/perf_fio.csv
         RemoteCopy -uploadTo $testVMData.PublicIP -port $testVMData.SSHPort -files ".\$constantsFile,.\$LogDir\StartFioTest.sh,.\$LogDir\ParseFioTestLogs.sh" -username "root" -password $password -upload
         RemoteCopy -uploadTo $testVMData.PublicIP -port $testVMData.SSHPort -files $currentTestData.files -username "root" -password $password -upload
         $out = RunLinuxCmd -ip $testVMData.PublicIP -port $testVMData.SSHPort -username "root" -password $password -command "chmod +x *.sh" -runAsSudo
-        $testJob = RunLinuxCmd -ip $testVMData.PublicIP -port $testVMData.SSHPort -username "root" -password $password -command "./StartFioTest.sh" -RunInBackground -runAsSudo
+        $testJob = RunLinuxCmd -ip $testVMData.PublicIP -port $testVMData.SSHPort -username "root" -password $password -command "bash StartFioTest.sh" -RunInBackground -runAsSudo
```
juhlee-microsoft: Are all distro support bash by default? this creates a new bash session to tun StartFioTest.sh vs. the previous line means it happens in the same session. It is different. (tip. IPC, folk)

Testscripts/Linux/perf_fio.sh
LiliDeng: We changed fio run against file into run against device directly. On azure, we can't see much differences on Azure, since results reaches limit. But it has better results on Hyper-V. Also for run against device, it no need to prepare file any more. Save much time.

Testscripts/Linux/perf_fio.sh
LiliDeng: @juhlee-microsoft Ok, I will replace them. Thanks for the advice.

Testscripts/Linux/perf_fio.sh
LiliDeng: @juhlee-microsoft run against device directly, no need make file system and mount any more.

Testscripts/Linux/perf_netperf.sh
LiliDeng: @juhlee-microsoft I move these line backwards. based my understanding, it is meaningful when client started. So I move monitoring server part after start client. Let know my understanding is not correct.

Testscripts/Windows/PERF-STORAGE-MULTIDISK-RAID0-FIO.ps1
LiliDeng: @juhlee-microsoft It is hard to explain here, because I didn't know the root cause. If I didn't add bash here, these two lines in file StartFioTest.sh won't execute against coreos.

. utils.sh
collect_VM_properties

lizzha: @juhlee-microsoft as far as I know, we haven't met a distro in our test that doesn't support bash

LiliDeng: @juhlee-microsoft I ran test against ubuntu (regression) and coreos (functional), and send results to you last week, I will run against RHEL today.

juhlee-microsoft: if you ran Ubuntu, then it would be ok without any error.

PR #323, opened 2018-10-26
hyper-v : Fix ETHTOOL-CHECK-STATISTICS 

asofron:
ETHTOOL-CHECK-STATISTICS testcase should not fail if **Wake_queue** value greater than 10

Before Fix :

Sr. Test Name : Test Result : Test Duration (in minutes)
----------------------------------------------------
1. ETHTOOL-CHECK-STATISTICS : ABORTED : 11.05

After Fix :

Sr. Test Name : Test Result : Test Duration (in minutes)
 ----------------------------------------------------
 1. ETHTOOL-CHECK-STATISTICS : PASS : 36.82


asofron:
Refactored and removed check

Testscripts/Linux/NET-Ethtool-Check-Statistics.sh
```diff
@@ -247,14 +247,6 @@ ChangeMTU(){
     #Log the values
     LogMsg "Wake_queue start value: $wake_value"
     LogMsg "Wake_queue value after changing MTU: $new_wake_value"
-
-    if [ $new_wake_value -eq 10 ]; then
```
juhlee-microsoft: This verification is still needed. If the chipset fails to wake up then it will escalate the failure.

chvalean: I've asked Andrei to make this change (same was done on lisav1) because the wake_queue pattern has changed since this automation was created, and we have a 100% test failure because of this.
Transforming the failure to a warning, we still do the rest of ethtool coverage that we need.

juhlee-microsoft: @chvalean I don't have full confidence of the change without team's review. You can switch to warning message, but I'd like you to make sure this is right change with kernel developers. Can you get their confirmation? It's could be mlx driver or kernel change so far.

chvalean: This has been discussed internally already

juhlee-microsoft: ok.

Issue #330, opened 2018-10-29
lisav2 fails when the ssh host key is no trusted


ader1990:
pscp and plink should always have a write-output "yes" | command

PR #334, opened 2018-10-30
Fix : lisav2 fails when the ssh host key is not trusted #330

asofron:
Added Write-Output "yes" , before pscp or plink in order update host key in case it changed


lizzha:
Please post test results

PR #340, opened 2018-10-30
lisav2: Don't copy temp files on workspace change

ader1990:
When lisav2 workspace is changed there is no need to have all the .git files copied plus a lot of temp files from previous runs, which can linearly increase the times for lisav2 to run, without manual cleanup.

PR tested internally and works fine.

### DIFF
--- a/Run-LisaV2.ps1
+++ b/Run-LisaV2.ps1
@@ -112,7 +112,10 @@ try {
 		$finalWorkingDirectory = "$TempWorkspace\LISAv2\$shortRandomWord$shortRandomNumber"
 		$tmpSource = '\\?\' + "$OriginalWorkingDirectory\*"
 		Write-Output "Copying current workspace to $finalWorkingDirectory"
-		Copy-Item -Path $tmpSource -Destination $finalWorkingDirectory -Recurse -Force | Out-Null
+		$excludedDirectories = @(".git", ".github", "TestResults",
+			"VHDs_Destination_Path", "*.zip", "report")
+		Copy-Item -Path $tmpSource -Destination $finalWorkingDirectory `
+			-Recurse -Force -Exclude $excludedDirectories | Out-Null
 		Set-Location -Path $finalWorkingDirectory | Out-Null
 		Write-Output "Working directory has been changed to $finalWorkingDirectory"
 		$WorkingDirectory = $finalWorkingDirectory


Run-LisaV2.ps1
```diff
@@ -112,7 +112,10 @@ try {
 		$finalWorkingDirectory = "$TempWorkspace\LISAv2\$shortRandomWord$shortRandomNumber"
 		$tmpSource = '\\?\' + "$OriginalWorkingDirectory\*"
 		Write-Output "Copying current workspace to $finalWorkingDirectory"
-		Copy-Item -Path $tmpSource -Destination $finalWorkingDirectory -Recurse -Force | Out-Null
+		$excludedDirectories = @(".git", ".github", "TestResults",
```
iamshital: We need TestResults, zip and report.

juhlee-microsoft:
@iamshital Can you clarify this is your design? Also I know there is windows limitation of length of file path. Is it the original purpose of file copying back/forth? why is not required today?

Run-LisaV2.ps1
```diff
@@ -112,7 +112,10 @@ try {
...
+			"VHDs_Destination_Path", "*.zip", "report")
+		Copy-Item -Path $tmpSource -Destination $finalWorkingDirectory `
```
juhlee-microsoft: there is also fily copying back statement in the same file. (There are three Copy-Item statements)

ader1990: the max path on some systems is 256 char long, otherwise you end up to have files that exist but are not seen by some applications. THIS feature is still required.

iamshital:
> @iamshital Can you clarify this is your design? Also I know there is windows limitation of length of file path. Is it the original purpose of file copying back/forth? why is not required today?

Hi Ju,

When we execute the LISAv2 using jenkins **Pipelines** (not jobs), I often observed that, no. of characters in working directory exceed 150-160 characters. When LISAv2 creates it's own files, sometimes filepath exceed 256 characters and LISAv2 fails to access existing files.

We workaround this using copying file to shorter working directory --> Execute the tests --> Copy back the results to original working directory.

@ader1990 is not removing the current flow. He is only excluding some folders which are not required for LISAv2 execution.

@juhlee-microsoft , let me know, if you still have any confusion.

PR #346, opened 2018-11-01
Check 'Status' to determine VM state & add retry

santoshx:
We dont need just start-vm command to succeed.
We need VMs to be in running state.

Without this my tests with three or more VMs fails at least once in three runs.


Libraries/HyperV.psm1
```diff
@@ -544,16 +544,30 @@ Function StartHyperVGroupVMs($HyperVGroupName,$HyperVHost)
     $CurrentErrors = @()
     foreach ( $VM in $AllVMs.VMMembers)
     {
+        $RetryCount = 5
+        $CurrentRetryAttempt=0
+        $res = "False"
```
lizzha: use $false, $true

Libraries/HyperV.psm1
```diff
@@ -544,16 +544,30 @@ Function StartHyperVGroupVMs($HyperVGroupName,$HyperVHost)
     $CurrentErrors = @()
     foreach ( $VM in $AllVMs.VMMembers)
     {
+        $RetryCount = 5
```
ader1990: why not use Wait-VMState method? it is more reliable

Libraries/HyperV.psm1
ader1990: also, what you are saying is that your vms do not start correctly? what is the reason for that?

usually vms don t start when they don t have enough resources like ram/cpu or badly configured hardware.

santoshx: After this your comment I observed failures for a day and found all failures were due to lack of resources. This PR is not needed. Thank you.

PR #349, opened 2018-11-01
azure: ARMImageName param can be empty when OSVhd is set

ader1990:


ader1990:
tested with osvhd param on azure and without ARMImage param, works as intended.

PR #351, opened 2018-11-02
Fix the error from PowerShellCodeCheck_PSScriptAnalyzer Job

LiliDeng:
I am testing if this PR can pass the check.


LiliDeng:
@juhlee-microsoft Could you please help check why upload files failed, seems the user didn't have the access.
I am running regression testing now, will update results later.

juhlee-microsoft:
> @juhlee-microsoft Could you please help check why upload files failed, seems the user didn't have the access.
> I am running regression testing now, will update results later.

I am not sure if this is from security issue or recent change. Many moving balls. I currently work on security issue fix. @chvalean  any comment?

PR #356, opened 2018-11-05
Add 2 & 3 VMs configuration for DM tests

santoshx:
Dynamic Memory tests with 2 and 3 VMs need this.


adriansuhov:
Why is this not part of the PR that will include the actual DM tests?

santoshx:
@adriansuhov more than one person working on DM tests to avoid duplicate work I added it here

PR #385, opened 2018-11-14
Implementation for VM creation on cluster

sireeshapotnuru:
This Implementation supports  VM creation on a cluster environment.


sireeshapotnuru:
Need to Pass -TiPCluster "True" when invoking Run-LisaV2.ps1

Libraries/Framework-Azure.psm1
```diff
@@ -102,6 +102,16 @@ function Validate-Parameters {
     } else {
         LogMsg "Test parameters have been validated successfully. Continue running the test."
     }
+    if ($TiPCluster -eq "True") {
```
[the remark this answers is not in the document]
chvalean: Probably defining the VM to be clustered would go as a VM flavor in the xml list of VM definitions?
Not sure to recommend the best option to do this.

Libraries/Framework-HyperV.psm1
```diff
@@ -589,6 +589,10 @@ function Run-Test {
     $currentTestResult.TestResult = GetFinalResultHeader -resultarr $resultArr
     if ($DeployVMPerEachTest -or $ExecuteTeardown) {
         LogMsg "VM CLEANUP ~~~~~~~~~~~~~~~~~~~~~~~"
+        foreach ($VM in $AllVMData) {
+            $HyperVGroup=$VM.HyperVGroupName
+            Add-VMGroupMember -Name "$HyperVGroup" -VM (Get-VM -name $VM.RoleName) -ComputerName $VM.HyperVHost
```
LiliDeng: Get-VM  add -ComputerName parameter?

Libraries/Framework-Azure.psm1
```diff
@@ -102,6 +102,16 @@ function Validate-Parameters {
...
+        Get-Cluster | out-null
```
LiliDeng: Is that possible you run cluster testing remotely? if yes, should add computername parameter?

Libraries/Framework-Azure.psm1
LiliDeng: You can make this parameter as [switch] type like EnableAcceleratedNetworking, then if ($TiPCluster) is enough to do judgement

Libraries/HyperV.psm1
```diff
@@ -477,6 +520,13 @@ Function CreateHyperVGroupDeployment([string]$HyperVGroup, $HyperVGroupNameXML,
                 $ErrorCount += 1
             }
         }
+        if($TiPCluster -eq "True") {
+            Move-VMStorage $CurrentVMName -DestinationStoragePath $DestinationOsVHDPath
```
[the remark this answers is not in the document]
sireeshapotnuru: Thanks Chris

Libraries/Framework-Azure.psm1
```diff
@@ -102,6 +102,16 @@ function Validate-Parameters {
...
+        if ($? -eq $False){
+            throw "Server doesn't have a cluster set up"
+        }
+        $clusterDir = Get-ClusterSharedVolume
+        if( -not $clusterDir) {
+            throw "No Cluster Shared Volume"
```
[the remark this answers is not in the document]
sireeshapotnuru: modified throw message

Libraries/Framework-HyperV.psm1
sireeshapotnuru: added -ComputerName parameter

Libraries/Framework-Azure.psm1
sireeshapotnuru: cant add computername  param to get-cluster command

Libraries/Framework-Azure.psm1
sireeshapotnuru: Used new switch parameter -ClusteredVM

sireeshapotnuru:
Now, we need to pass -ClusteredVM when invoking Run-LisaV2.ps1

AutomationManager.ps1
```diff
@@ -90,6 +91,10 @@ if($EnableAcceleratedNetworking) {
 	Set-Variable -Name EnableAcceleratedNetworking -Value $true -Scope Global
 }
 
+if($ClusteredVM) {
+	Set-Variable -Name ClusteredVM -Value $true -Scope Global
```
chvalean: Simon recommend that we should not introduce any new global variables.
Please look into that, or if the effort is too much maybe @juhlee-microsoft  can confirm an exception.

ader1990:
Hello,

This implementation seems pretty simple and might not catch all the issues when a live migration is tested. For this change it is mandatory to add an example test case, so that we can discuss on how this change can be better fitted to the needs of a real test case.

What are the requirements of a test case run in LISAV2 with a clustered VM?

As I can see, there cannot be a LISAV2 run with test cases that have clustered or non clustered VMs.

For start, can you add a test case that uses this framework change?

Thank you,
Adrian

ader1990:
@sireeshapotnuru can you add the SECUREBOOT-MIGRATEVM code in the same PR?

AutomationManager.ps1
ader1990: I think this clusteredvm should be a setuptype variable

Libraries/HyperV.psm1
```diff
@@ -67,6 +67,12 @@ Function DeployHyperVGroups ($xmlConfig, $setupType, $Distro, $getLogsIfFailed =
                         $retValue = $NULL
                     }
                 }
+                if($ClusteredVM) {
```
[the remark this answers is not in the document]
sireeshapotnuru: There is a lot of dependency for the vm to be in vmgroup, functions StartHyperVGroupVMs, GetAllHyperVDeployementData, uses HyperVGroupName to start vm & for setting $allVMData

sireeshapotnuru:
@ader1990, I have added the Testcase "SECUREBOOT-MIGRATEVM" in this PR

AutomationManager.ps1
sireeshapotnuru: @ader1990 do you want me need to add a new setuptype <ClusteredVM> with <isClusteredVM>True</isClusteredVM> and check using value of 
$xmlConfig.config.HyperV.Deployment.$setupType.isClusteredVM everywhere?

sireeshapotnuru:
Added new setuptype "ClusteredVM"

Libraries/HyperV.psm1
```diff
@@ -114,7 +120,12 @@ Function CreateAllHyperVGroupDeployments($setupType, $xmlConfig, $Distro, $Debug
             }
 
             $SourceOsVHDPath = $xmlConfig.config.HyperV.Hosts.ChildNodes[$index].SourceOsVHDPath
-            $DestinationOsVHDPath = $xmlConfig.config.HyperV.Hosts.ChildNodes[$index].DestinationOsVHDPath
+            if ($setupType -eq "ClusteredVM") {
+                $clusterDir = Get-ClusterSharedVolume
```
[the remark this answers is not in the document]
ader1990: in this case, Run-LisaV2 base path is not used, this has to be documented

Libraries/HyperV.psm1
```diff
@@ -281,6 +292,35 @@ Function DeleteHyperVGroup([string]$HyperVGroupName, [string]$HyperVHost) {
         }
         Remove-VM -Name $vm.Name -ComputerName $HyperVHost -Force
         LogMsg "Hyper-V VM $($vm.Name) removed!"
+        if ($CurrentTestData.setupType -eq "ClusteredVM") {
+            LogMsg "Deleting VM on Cluster"
+            Get-Command "Get-ClusterResource" -ErrorAction SilentlyContinue
+            if ($?) {
+                $group = Get-ClusterGroup -ErrorAction SilentlyContinue
+                $vm_matches = $group -match $vm.Name
```
ader1990: be consistent, don t use bash like camel_case naming code in PowerShell

Libraries/HyperV.psm1
```diff
@@ -336,7 +376,12 @@ Function CreateHyperVGroupDeployment([string]$HyperVGroup, $HyperVGroupNameXML,
                 $hostNumber = $HyperVGroupXML.VirtualMachine.indexOf($VirtualMachine)
                 $HyperVHost = $xmlConfig.config.HyperV.Hosts.ChildNodes[$hostNumber].ServerName
                 $SourceOsVHDPath = $xmlConfig.config.HyperV.Hosts.ChildNodes[$hostNumber].SourceOsVHDPath
-                $DestinationOsVHDPath = $xmlConfig.config.HyperV.Hosts.ChildNodes[$hostNumber].DestinationOsVHDPath
+                if ($CurrentTestData.setupType -eq "ClusteredVM") {
```
ader1990: ditto

Libraries/HyperV.psm1
```diff
@@ -402,6 +447,7 @@ Function CreateHyperVGroupDeployment([string]$HyperVGroup, $HyperVGroupNameXML,
                 LogMsg "New-VM -Name $CurrentVMName -MemoryStartupBytes $CurrentVMMemory -BootDevice VHD -VHDPath $CurrentVMOsVHDPath -Generation $VMGeneration -Switch $($VMSwitches.Name) -ComputerName $HyperVHost"
                 $NewVM = New-VM -Name $CurrentVMName -MemoryStartupBytes $CurrentVMMemory -BootDevice VHD `
                     -VHDPath $CurrentVMOsVHDPath -Generation $VMGeneration -Switch $($VMSwitches.Name) -ComputerName $HyperVHost
+                Enable-VMIntegrationService -Name "Guest Service Interface" -vmName $CurrentVMName -ComputerName $HyperVHost
```
ader1990: why is this needed, all the integration services should have been already enabled

Libraries/HyperV.psm1
```diff
@@ -461,6 +507,13 @@ Function CreateHyperVGroupDeployment([string]$HyperVGroup, $HyperVGroupNameXML,
                 LogMsg "Prerequiste: Prepare OS Disk $CurrentVMOsVHDPath - Failed."
                 $ErrorCount += 1
             }
+            if($CurrentTestData.setupType -eq "ClusteredVM") {
+                Move-VMStorage $CurrentVMName -DestinationStoragePath $DestinationOsVHDPath
+                Add-ClusterVirtualMachineRole -VirtualMachine $CurrentVMName
+                if ($? -eq $False) {
+                    throw "High Availability configure for ${CurrentVMName} failed. The VM was not created in cluster"
```
ader1990: throw command does not work in this scenario, revisit how the code works above and use the $ErrorCount

XML/TestCases/FunctionalTests.xml
```diff
@@ -2785,4 +2785,18 @@
         <Area>STORAGE</Area>
         <Tags>storage</Tags>
     </test>
+    <test>
+        <testName>SECUREBOOT-MIGRATEVM</testName>
+        <setupScript>.\TestScripts\Windows\SET-SECUREBOOT.ps1</setupScript>
+        <testScript>SECUREBOOT.ps1</testScript>
+        <Timeout>900</Timeout>
+        <setupType>ClusteredVM</setupType>
```
ader1990: we were discussing to put the clusteredvm setuptype to the vm setuptype, and not the test case one.
is there a reason you put it here?

sireeshapotnuru: we have introduced a new vm setuptype as "ClusteredVM" , which can be used for only cluster migration testcases and added the setuptype "ClusteredVM" in FunctionalTestsConfigurations.xml

Libraries/Framework-HyperV.psm1
```diff
@@ -581,6 +581,12 @@ function Run-Test {
 
     $currentTestResult.TestResult = GetFinalResultHeader -resultarr $resultArr
     LogMsg "VM CLEANUP ~~~~~~~~~~~~~~~~~~~~~~~"
+    if($CurrentTestData.setupType -eq "ClusteredVM") {
```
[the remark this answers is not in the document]
sireeshapotnuru: changed

Libraries/Framework-HyperV.psm1
```diff
@@ -581,6 +581,12 @@ function Run-Test {
...
+        foreach ($VM in $AllVMData) {
+            $HyperVGroup=$VM.HyperVGroupName
```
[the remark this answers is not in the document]
sireeshapotnuru: removed this variable and used $VM.HyperVGroupName

Libraries/HyperV.psm1
```diff
@@ -60,6 +60,12 @@ Function DeployHyperVGroups ($xmlConfig, $setupType, $Distro, $getLogsIfFailed =
                     $retValue = $NULL
                 }
             }
+            if($setupType -eq "ClusteredVM") {
```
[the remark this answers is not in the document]
sireeshapotnuru: broken into two lines

Libraries/HyperV.psm1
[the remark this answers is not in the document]
sireeshapotnuru: renamed to ClusterVolume

Libraries/HyperV.psm1
sireeshapotnuru: removed this line

Libraries/HyperV.psm1
sireeshapotnuru: modified variable name to MatchedVM

Libraries/Framework-HyperV.psm1
```diff
@@ -581,6 +581,12 @@ function Run-Test {
...
+            Add-VMGroupMember -Name "$HyperVGroup" -VM (Get-VM -name $VM.RoleName -ComputerName $VM.HyperVHost) -ComputerName $VM.HyperVHost
```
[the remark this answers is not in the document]
sireeshapotnuru: broken into two lines

Libraries/HyperV.psm1
sireeshapotnuru: changed

sireeshapotnuru:
Latest Test Result:

[LISAv2 Test Results Summary]
Test Run On           : 11/22/2018 10:22:40
VHD Under Test        : SecbootGen2.vhdx
Total Time (dd:hh:mm) : 0:0:4

    1 SECUREBOOT-MIGRATEVM                                               PASS                 4.27

XML/VMConfigurations/FunctionalTestsConfigurations.xml
```diff
@@ -682,4 +682,22 @@
             </VirtualMachine>
         </ResourceGroup>
     </ThreeVMsDM>
+    <ClusteredVM>
```
adriansuhov: Can this "ClusteredVM" tag be inside the SetupType? Let's take this scenario: If in the future we'll have a test case that requires 2 clustered VMs, then the framework should be changed. That means that we're stuck with only this setupType for a clustered VM(s) scenario. I would like to see a tag inside the Setuptype => this way we can reuse the clustered VMs logic to any SetupType.

sireeshapotnuru: As suggested added new setup "ClusteredSingleVM" with tag `<ClusteredVM>True</ClusteredVM>` and checking with $xmlConfig.config.HyperV.Deployment.($CurrentTestData.setupType).ClusteredVM

Libraries/Framework-HyperV.psm1
```diff
@@ -581,6 +581,12 @@ function Run-Test {
 
     $currentTestResult.TestResult = GetFinalResultHeader -resultarr $resultArr
     LogMsg "VM CLEANUP ~~~~~~~~~~~~~~~~~~~~~~~"
+    if ($xmlConfig.config.HyperV.Deployment.($CurrentTestData.setupType).ClusteredVM) {
+        foreach ($VM in $AllVMData) {
+            Add-VMGroupMember -Name $VM.HyperVGroupName -VM (Get-VM -name $VM.RoleName -ComputerName $VM.HyperVHost) `
+             -ComputerName $VM.HyperVHost
```
ader1990: leave a tab or 4 spaces for the indent

ader1990:
> * to another node irrespective of migration type
> * Should be able to stop the Cluster service on a node in

Hello,

I was referring to lisav2 running test cases using clustered vms.
I have some questions:
1. What are the requirements of the environment in addition to the hyperv lisav2 requirements?
2. Can I run other test cases in the same lisav2 run?

Please answer to all these questions in the commit message too.

Libraries/Framework-HyperV.psm1
```diff
@@ -581,6 +581,12 @@ function Run-Test {
 
     $currentTestResult.TestResult = GetFinalResultHeader -resultarr $resultArr
     LogMsg "VM CLEANUP ~~~~~~~~~~~~~~~~~~~~~~~"
+    if ($xmlConfig.config.HyperV.Deployment.($CurrentTestData.setupType).ClusteredVM) {
```
ader1990: what do you mean $currenttestdata.setuptype?

ader1990: $CurrentTestData.setupType is the actual VM setup type?

Libraries/HyperV.psm1
```diff
@@ -281,6 +291,35 @@ Function DeleteHyperVGroup([string]$HyperVGroupName, [string]$HyperVHost) {
...
+                $MatchedVM = $group -match $vm.Name
+                foreach ($vm in $MatchedVM) {
+                    Remove-ClusterGroup -Name $vm.name -RemoveResources -Force
+                    if (-not $?) {
+                        LogErr "Failed to remove Cluster Role for VM $vm"
+                        return $False
+                    }
+                    LogMsg "Cleanup was successful for $vm on Cluster"
+                }
+                # Also remove VM from second node if it's located there
```
ader1990: don t rely on the fact the there is only one other cluster node, remove the vm from all nodes.
this is a framework change and changes here should be generic, even if the test cases rely on the fact that there is a setup with only two cluster nodes

Libraries/HyperV.psm1
```diff
@@ -461,6 +505,14 @@ Function CreateHyperVGroupDeployment([string]$HyperVGroup, $HyperVGroupNameXML,
                 LogMsg "Prerequiste: Prepare OS Disk $CurrentVMOsVHDPath - Failed."
                 $ErrorCount += 1
             }
+            if ($xmlConfig.config.HyperV.Deployment.($CurrentTestData.setupType).ClusteredVM) {
+                Move-VMStorage $CurrentVMName -DestinationStoragePath $DestinationOsVHDPath
+                Add-ClusterVirtualMachineRole -VirtualMachine $CurrentVMName
+                if ($? -eq $False) {
+                    LogErr "High Availability configure for ${CurrentVMName} failed. The VM was not created in cluster"
```
ader1990: put something like
"VM ${CurrentVMName} could not be added to the Hyper-V cluster."

Libraries/Framework-HyperV.psm1
sireeshapotnuru: Yes

sireeshapotnuru:
> > * to another node irrespective of migration type
> > * Should be able to stop the Cluster service on a node in
> 
> Hello,
> 
> I was referring to lisav2 running test cases using clustered vms.
> I have some questions:
> 
> 1. What are the requirements of the environment in addition to the hyperv lisav2 requirements?
> 2. Can I run other test cases in the same lisav2 run?
> 
> Please answer to all these questions in the commit message too.

Answers as below: 
What are the requirements of the environment in addition to the hyperv lisav2 requirements?
	1) Should have a Cluster setup with atleast two nodes and one storage volume

Can I run other test cases in the same lisav2 run?
	2) yes we can run both Non-clustered and Clustered testcases in same run

Updated the same in commit message too

Libraries/Framework-HyperV.psm1
sireeshapotnuru: added tab

Libraries/HyperV.psm1
sireeshapotnuru: modified logerr line

Libraries/HyperV.psm1
sireeshapotnuru: changed logic which checks vm on other nodes and removes it

PR #387, opened 2018-11-15
Refine LISAv2 framework code

simonxiaoss:
1. Remove useless functions and logging
2. Refine the variable names


AutomationManager.ps1
```diff
@@ -134,16 +134,6 @@ if ($UseManagedDisks) {
 	Set-Variable -Name UseManagedDisks -Value $false -Scope Global
 }
 
-if ( $XMLSecretFile ) {
```
juhlee-microsoft: Because no $xmlSecrets variable leads subscription ID fetching failure, this needs to be back.

chvalean:
@simonxiaoss in the usual review process this needs to also include the results for sample lisav2 run for both Azure and Hyper-V.

AutomationManager.ps1
simonxiaoss: fixed.

PR #398, opened 2018-11-16
Cleanup the framework code

simonxiaoss:


.gitignore
```diff
@@ -7,4 +7,7 @@ UploadStatusFile*
 *.azure.env
 *-buildlogs.zip
 .vscode/settings.json
-*.exe
\ No newline at end of file
+*.exe
+Tools/kvp_*
```
juhlee-microsoft: Those are required binaries. Any reason you put ignore list?

Libraries/Azure.psm1
```diff
@@ -583,6 +585,7 @@ Function CreateResourceGroup([string]$RGName, $location) {
         }
         catch {
             $retValue = $false
+            LogErr "Exception in CreateResourceGroup"
```
juhlee-microsoft: There is recommended programming style in catch block. Please add 'Exception reason and line number'. You can find the working code from everywhere.

Libraries/Azure.psm1
```diff
@@ -637,6 +640,7 @@ Function CreateResourceGroupDeployment([string]$RGName, $location, $setupType, $
         }
         catch {
             $retValue = $false
+            LogErr "Exception in CreateResourceGroupDeployment"
```
juhlee-microsoft: The same comment applied for exceptional reason and line number.

.gitignore
simonxiaoss: Yes they are required by the test; but the git repo does not host them, so ignore.

Libraries/Azure.psm1
simonxiaoss: fixed; but need future rework to refine the exception handling

Libraries/Azure.psm1
simonxiaoss: fixed; but need future rework to refine the exception handling

PR #409, opened 2018-11-20
Refactor the code (XML processing)

simonxiaoss:
1. Put XML processing code into a PS module so that the main script looks more clear.
2. Other code style change and naming change.

Test passed on Azure.


AzureTestSuite.ps1
```diff
@@ -85,12 +85,15 @@ Function RunTestsOnCycle ($cycleName , $xmlConfig, $Distro, $TestIterations, $de
...
 
 	# Start JUnit XML report logger.
-	$reportFolder = "$pwd/report"
+	$reportFolder = "$pwd/Report"
 	if(!(Test-Path $reportFolder)) {
 		New-Item -ItemType "Directory" $reportFolder
 	}
 
-	StartLogReport("$reportFolder/report_$($testCycle.cycleName).xml")
+	$TestReportXml = Join-Path "$pwd" ".\Report\LISAv2_TestReport_$TestID.xml"
```
lizzha: Join-Path “$reportFolder” "LISAv2_TestReport_$TestID.xml" ?

Libraries/CommonFunctions.psm1
```diff
@@ -2492,18 +2468,20 @@ Function GetFilePathsFromLinuxFolder ([string]$folderToSearch, $IpAddress, $SSHP
 	return $LogFilesPaths, $LogFiles
 }
 
-function ZipFiles( $zipfilename, $sourcedir )
+function Zip-Files( $zipFileName, $sourceDir )
```
lizzha: better to name it New-ZipFile

LiliDeng:
@simonxiaoss Bug find when run cases against hyperv, fixing now.

PR #411, opened 2018-11-20
azure: get kernel logs in case of boot fail

ader1990:


Libraries/Azure.psm1
```diff
@@ -1960,36 +1960,42 @@ Function isAllSSHPortsEnabledRG($AllVMDataObject) {
...
+        } else {
             LogMsg "ALL VM's port $port is/are open now.."
             $retValue = "True"
         }
 
+    } While (($timeout -lt 20) -and ($WaitingForConnect -gt 0))
+
+	if ($retValue -eq "False") {
+		foreach ($vm in $AllVMDataObject) {
+			$out = Test-TCP -testIP $($vm.PublicIP) -testport $port
```
simonxiaoss: This is good; can we put this as a Function, for example:
Get-AzureVmSerialLog($vm)

ader1990: sure, will be done

LiliDeng:
@ader1990 where is the definition of GetcheckConnectivityFile(), it will hit issue when create $initialBootLogFile, line 760 of CommonFunctions.psm1
 @juhlee-microsoft @simonxiaoss

Libraries/CommonFunctions.psm1
```diff
@@ -747,126 +747,115 @@ Function WaitFor($seconds,$minutes,$hours)
...
+			$currenBootLog = $initialBootLog
+			$kernelLogStatus = Join-Path $BootLogDir "KernelLogStatus.txt"
+
+			if ($status -imatch "Final") {
+				$currenBootLogFile = $finalBootLogFile
+				$currenBootLog = $finalBootLog
+			}
+
+			if ($status -imatch "Initial") {
+				$checkConnectivityFile = Join-Path $LogDir ([System.IO.Path]::GetcheckConnectivityFile())
```
LiliDeng: @ader1990 I didn't see this function GetcheckConnectivityFile definition.

lizzha: @ader1990 please fix this, and also address Simon's comment, and come out a new PR.

PR #420, opened 2018-11-21
kernel_logs: fix initial log file

ader1990:
The current log file should be the initial log file, not the final log file.
This PR fixes the bad implementation.


ader1990:
The current log file should be the initial log file, not the final log file.
This PR fixes the bad implementation.

PR #421, opened 2018-11-21
code cleanup: remove TC_COVERED references

chvalean:
automation no longer requires nor makes use of this parameter, removing it from the remaining files.


juhlee-microsoft:
@chvalean Changes are ok but please post the result of those changed test cases.

chvalean:
> @chvalean Changes are ok but please post the result of those changed test cases.

This is cleanup, no testing required as all references for those lines have been removed.

PR #423, opened 2018-11-22
Add a parameter TestPriority, to enable run test by priority

xian123:
1 Fix the bug of update the value of TestCategory,TestArea,TestNames and TestTag failed
2 Add a parameter TestPriority, when it is enabled, cases exact equal priorities will be run

### DIFF
--- a/Libraries/XmlProcessing.psm1
+++ b/Libraries/XmlProcessing.psm1
@@ -29,10 +29,6 @@ Function Import-TestCases($WorkingDirectory, $TestConfigurationXmlFile) {
 	$TestXMLs = Get-ChildItem -Path "$WorkingDirectory\XML\TestCases\*.xml"
 	$SetupTypeXMLs = Get-ChildItem -Path "$WorkingDirectory\XML\VMConfigurations\*.xml"
 	$AllLisaTests = @()
-	if ( $TestCategory -eq "All") { $TestCategory = "" }
-	if ( $TestArea -eq "All") {	$TestArea = "" }
-	if ( $TestNames -eq "All") { $TestNames = "" }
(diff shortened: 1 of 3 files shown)


lizzha:
@xian123 please resolve the conflict

Libraries/XmlProcessing.psm1
```diff
@@ -30,10 +30,6 @@ Function Import-TestCases($WorkingDirectory, $TestConfigurationXmlFile) {
 	$SetupTypeXMLs = Get-ChildItem -Path "$WorkingDirectory\XML\VMConfigurations\*.xml"
 	$AllLisaTests = @()
 	$ARMImage = $ARMImageName.Trim().Split(" ")
-	if ( $TestCategory -eq "All") { $TestCategory = "" }
```
lizzha: These are changed in another refactor PR, please double check where to place these line is better

lizzha:
@ader1990 it's required by Simon to enable running test by test priority. 
Now we have most test cases with priority defined. Chris will add priority for the rest that are imported later.

Libraries/XmlProcessing.psm1
xian123: Actually, the $TestCategory, $TestArea, etc. are **global** variables and they are not changed the **global** value just by $varName = value format.   That is, their values are "All" not null in CollectTestCases() if they are changed values out of CollectTestCases() via $varName = value.

LiliDeng:
@ader1990 @lizzha Thanks for review, any other comments?

juhlee-microsoft:
@xian123 do you have test result with TestPriority? I am excited to seeing it.

Issue #438, opened 2018-11-26
Hyper-V: VM handling in case of failures


chvalean:
Also covering scenarios where SetupScript requires deploy of a new VM.

Current workflow is that in case of test failure the VM will be left running, and also having a different VM size will deploy another VM on Hyper-V.
This leaves the old VM running, leading to no available resources.

Changes discussed will involve:
Test case exists with ABORT/FAIL -> create VM snapshot -> VM Forced turnOff
SetupScript to deploy a different VM -> create VM snapshot -> VM Forced turnOff

Will create a new issue to handle VM cleanup after several days, similar with the Azure cleanup.

PR #442, opened 2018-11-26
hyperv: optimize resource consumption

ader1990:
Addresses: https://github.com/LIS/LISAv2/issues/438


ader1990:
In order to improve resource consumption on lisav2 HYPERV,
in case of a failed or aborted test, we need to create a checkpoint
and stop the VM so that it will not consume RAM/CPU on the test
environment.

This is necessary as the lisav2 hyperv test vms are created once a
new setup type is found and there are almost 200 test cases with at
least 10 setup types, and if some of the test cases are not passed,
the next VMs will not be able to start mainly due to out of RAM
exception.

ader1990:
@juhlee-microsoft the answer is simple math, as the new behavior is to stop VMs after the test cases finish with an abort/failure.
The previous behavior was to leave the VMs running. The resources depend on the test case setup type.

Libraries/HyperV.psm1
```diff
@@ -252,12 +252,19 @@ Function DeleteHyperVGroup([string]$HyperVGroupName, [string]$HyperVHost) {
         return $true
     }
 
+    $cleanupDone = 0
     $vmGroup.VMMembers | ForEach-Object {
-        LogMsg "Stop-VM -Name $($_.Name) -Force -TurnOff "
         $vm = $_
+        LogMsg "Stop-VM -Name $vm -Force -TurnOff"
         Stop-VM -Name $vm.Name -Force -TurnOff -ComputerName $HyperVHost
+        $snapshots = Get-VMSnapshot -VMName $vm.Name
```
LiliDeng: Get-VMSnapshot -VMName $vm.Name -ComputerName $HyperVHost

Libraries/HyperV.psm1
```diff
@@ -322,9 +329,11 @@ Function DeleteHyperVGroup([string]$HyperVGroupName, [string]$HyperVHost) {
         }
     }
 
-    LogMsg "Hyper-V VM group ${HyperVGroupName} is being removed!"
-    Remove-VMGroup -Name $HyperVGroupName -ComputerName $HyperVHost -Force
-    LogMsg "Hyper-V VM group ${HyperVGroupName} removed!"
+    if ($cleanupDone) {
+        LogMsg "Hyper-V VM group ${HyperVGroupName} is being removed!"
+        Remove-VMGroup -Name $HyperVGroupName -ComputerName $HyperVHost -Force
+        LogMsg "Hyper-V VM group ${HyperVGroupName} removed!"
```
LiliDeng: It will be executed when the VM has more than 1 checkpoints, in this situation, the group is not empty, then throw exception.

Remove-VMGroup : The operation failed.
An unexpected error occurred: The library, drive, or media pool must be empty to perform this operation. (0x800710D3).
The operation failed.
The Hyper-V Virtual Machine Management service encountered an unexpected error: The library, drive, or media pool must be empty to perform this operation. 
(0x800710D3).
At E:\lili\ader1990\test\Libraries\HyperV.psm1:334 char:9
+         Remove-VMGroup -Name $HyperVGroupName -ComputerName $HyperVHo ...
+         ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : NotSpecified: (:) [Remove-VMGroup], VirtualizationException
    + FullyQualifiedErrorId : OperationFailed,Microsoft.HyperV.PowerShell.Commands.RemoveVMGroup

Libraries/HyperV.psm1
```diff
@@ -252,12 +252,19 @@ Function DeleteHyperVGroup([string]$HyperVGroupName, [string]$HyperVHost) {
...
     }
 
+    $cleanupDone = 0
     $vmGroup.VMMembers | ForEach-Object {
-        LogMsg "Stop-VM -Name $($_.Name) -Force -TurnOff "
         $vm = $_
+        LogMsg "Stop-VM -Name $vm -Force -TurnOff"
         Stop-VM -Name $vm.Name -Force -TurnOff -ComputerName $HyperVHost
+        $snapshots = Get-VMSnapshot -VMName $vm.Name
+        if ($snapshots.Count -gt 1) {
```
LiliDeng: Get-VMSnapshot -VMName $vm.Name -ComputerName $HyperVHost

Libraries/HyperV.psm1
```diff
@@ -925,3 +934,123 @@ Function Get-VMDemandMemory {
...
+        [boolean] $ShouldTurnOnVMAfterCheckpoint = $true,
+        [string] $CheckpointType = "Standard"
+    )
+
+    foreach ($VM in $VMData) {
+        if ($ShouldTurnOffVMBeforeCheckpoint) {
+            Stop-VM -Name $VM.RoleName -ComputerName $VM.HyperVHost -TurnOff -Force
+        } else {
+            Start-VM -Name $VM.RoleName -ComputerName $VM.HyperVHost
+        }
```
LiliDeng: $ShouldTurnOffVMBeforeCheckpoint you stop the vm, if not set, why start vm?

Libraries/LISAV2-Framework.psm1
```diff
@@ -0,0 +1,1781 @@
+##############################################################################################
+# Framework-Azure.psm1
```
LiliDeng: LISAV2-Framework.psm1

Libraries/WorkingSpaceManagement.psm1
```diff
@@ -28,29 +28,31 @@
...
 	$tmpDest = '\\?\' + $OriginalWorkingDirectory
 	Copy-Item -Path "$currentWorkingDirectory\*" -Destination $tmpDest -Force -Recurse | Out-Null
 	Set-Location ..
-	Write-Host "Cleaning up $currentWorkingDirectory"
-	Remove-Item -Path $currentWorkingDirectory -Force -Recurse -ErrorAction SilentlyContinue
-	Write-Host "Setting workspace back to original location: $originalWorkingDirectory"
+	if (!$ExitCode) {
+		LogMsg "Cleaning up $currentWorkingDirectory"
+		Remove-Item -Path $currentWorkingDirectory -Force -Recurse -ErrorAction SilentlyContinue
+	}
```
LiliDeng: Whenever if pass or fail, it will copy everything back, why need keep currentWorkingDirectory?

ader1990: when it fails, the backing vhd files for the VMs will be there, so it must not remove them.

Libraries/HyperV.psm1
ader1990: this is an enforcing parameter, if it does not turn off the VM, it will turn the vm on.

Libraries/HyperV.psm1
ader1990: outdated code, this issue has been solved

PR #444, opened 2018-11-27
No need enable root for testing on Azure.

LiliDeng:
Run case against Azure and HyperV pass (TestCategory BVT)


chvalean:
Common tests for Azure and Hyper-V or even standalone ones for Hyper-V only would have the same behavior - either require root or not.
So I don't think the difference to enable root user for Hyper-V only is something applicable per platform.
Or do you have any specific examples?

LiliDeng:
Please hold on this PR, I want to run perf testing to make sure not break them.

juhlee-microsoft:
I found your msg later, if you see any perf issue, please open a bug.

chvalean:
@juhlee-microsoft  this should not have been merged based on my question to Lili.
I don't understand why there is a Hyper-V inclusion only to run Enable-RootUser.
Independent on the platform - Azure or Hyper-V, triggering Enable-RootUser should not be required on one or the other.

juhlee-microsoft:
Oops. I missed it. Will look into this afternoon and find way to fix.


________________________________

PR #450, opened 2018-11-28
Revert "No need enable root for testing on Azure. (#444)"

juhlee-microsoft:
This reverts commit 56db225b5cbae93b5cc46138cb7f6b4bea479429.
Perf measurement and Hyper-V clarification require.


LiliDeng:
@juhlee-microsoft I run a performance case without issue, I don't understand the question. @chvalean 
In our legacy code [https://github.com/Azure/azure-linux-automation], we don't enable root for testing, unless some cases needs this situation.
There is a case BVT-IS-ROOT-PASSWORD-DELETED which check the distro vendor disable root user by default, but we enable root before starting testing, so this case will fail.
We think no need enable root for each case, unless your testing require this, like run ntttcp etc.

Issue #201, opened 2018-09-11
Remove LinuxRelease() in provisionLinuxForLisa.sh


juhlee-microsoft:
Replace with GetDistro from utils.sh

chvalean:
@juhlee-microsoft @iamshital I think we just have to remove the entire script provisionLinuxForLisa.sh
It is no longer required as tests will be imported in lisav2, so there is no VM preparation required for lisa.

Let's just do that and close this item.

lizzha:
this script is used in function ProvisionVMsForLisa, it enables root user and creates/upload same SSH key to client and server, which is required by network performance cases. It's still needed.
It's better to change a function/script name.

juhlee-microsoft:
According to @lizzha, we still need to change. Please keep this item in backlog.

chvalean:
I verified that function and although the provisionLinuxForLisa.sh script is copied to the VM, I don't see it getting executed from the function code.
@iamshital  can you please verify the code and maybe do some cleanup if provisionLinuxForLisa.sh is not actually used?

lizzha:
@chvalean you're right, the script is not executed. Only enableRoot.sh and enablePasswordLessRoot.sh are executed. So it can be removed.
Still we'd better change a function name of ProvisionVMsForLisa.

PR #453, opened 2018-11-28
fix issue https://github.com/LIS/LISAv2/issues/201

v-sijliu:
Pick up 3  BVT cases against Azure - Pass
Pick up 1 cases run against Hyper-V - Pass


LiliDeng:
You need check more against different distro.

Issue #452, opened 2018-11-28
Now all log go into AddAzureRmAccountFromSecretsFile.log


LiliDeng:
https://github.com/LIS/LISAv2/pull/445 This PR cause this issue. @chvalean

chvalean:
Thank you @LiliDeng 
This has been fixed now in https://github.com/LIS/LISAv2/pull/457

PR #457, opened 2018-11-28
Fix issue #452

asofron:
If main run-lisav2 script will call these scripts, it will conflict in using the $LogFileName variable.
This is to fix that behavior if the scripts are ran independently of lisav2


LiliDeng:
@chvalean @asofron Please provide test results for future PR, make sure not break case against HyperV and Azure.

asofron:
@LiliDeng  Update with relevant tests

Issue #467, opened 2018-11-29
Move STRESSTEST-RELOAD-MODULES-UP from BVT suite to Stress


ader1990:
STRESSTEST-RELOAD-MODULES-UP test case is in the BVT category and takes more than 100 minutes to complete. The whole Azure BVT suite takes around 3 hours, from which half the time is consumed by this test. The name of this test implies it should be in the BVT category.

I think this test case should be moved to stress category.
This test satisfies the stress test paradigm, please check https://en.wikipedia.org/wiki/Stress_testing_(software)

PR #471, opened 2018-11-29
Fix duplicate testcase and address issue #467

asofron:
- Removed CORE-VERIFY-LIS-MODULES-VERSION ( and test script). This exists already in BVT - BVT-CORE-VERIFY-LIS-MODULES.
- Move STRESSTEST-RELOAD-MODULES-UP from BVT to Stress category, to reduce runtime duration for BVT.


XML/TestCases/BVT.xml
```diff
@@ -245,22 +245,6 @@
         <Tags>core</Tags>
         <Priority>0</Priority>
     </test>
-    <test>
```
ader1990: this should not be deleted.

only put stress category here: https://github.com/LIS/LISAv2/blob/master/XML/TestCases/StressTests.xml#L43

asofron: Fixed

PR #491, opened 2018-12-04
perf_framework: iperf3: output perf results in a generic way 

ader1990:
To compare the results for the performance testing, we need a framework that:

- is generic - should work for all performance results
- has machine readable output

For genericity, all performance test cases should have a JSON output, with the name 

> TestCaseName_perf_results.json

The JSON file should contain an array of objects, with the object having the following structure
{
  "Id": ValueID,
 "ResultType1": Value1,
 "ResultType2": Value2,
 "ResultType3": Value3
}

Is mandatory to have at least the Id field.

This has to be implemented for all performance test cases to be able to have a comparison of the results.
If the test does not implement, there will be no result comparison possible.

The changes in the other test cases should require refactoring and adding this feature.
This feature is backwards compatible with the previous workflow.

### DIFF
--- a/Testscripts/Windows/PERF-NETWORK-TCP-SINGLE-CONNECTION-THROUGHPUT.ps1
+++ b/Testscripts/Windows/PERF-NETWORK-TCP-SINGLE-CONNECTION-THROUGHPUT.ps1
@@ -4,54 +4,151 @@ param(
     [String] $TestParams
 )
 
+function Write-TestInformation {
+    param(
+        $clientVMData,
+        $serverVMData
+    )
+
+    Write-LogInfo "CLIENT VM details :"
+    Write-LogInfo "  RoleName  : $($clientVMData.RoleName)"
+    Write-LogInfo "  Public IP : $($clientVMData.PublicIP)"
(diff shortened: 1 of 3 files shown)


ader1990:
Execution tested:
[azure_perf] [LISAv2 Test Results Summary]
[azure_perf] Test Run On           : 12/05/2018 15:16:06
[azure_perf] ARM Image Under Test  : Canonical : UbuntuServer : 18.04-DAILY-LTS : latest
[azure_perf] Total Test Cases      : 1 (1 Pass, 0 Fail, 0 Abort)
[azure_perf] Total Time (dd:hh:mm) : 0:0:10
[azure_perf] XML File              : TestConfiguration
[azure_perf] 
[azure_perf]    ID TestCaseName                                                 TestResult TestDuration(in minutes) 
[azure_perf] ------------------------------------------------------------------------------------------------------
[azure_perf]     1 PERF-NETWORK-TCP-SINGLE-CONNECTION-THROUGHPUT-SYNTHETIC                 PASS                10.69 
[azure_perf] 
[azure_perf]

PR #502, opened 2018-12-06
stor_perf: fix backwards compatibility in params

ader1990:
The storage bash scripts require arrays to be defined as:
'val1 val2'

and not:
(val1 val2)

This patch fixes perf_fio.sh to manage arrays given as (val1 val2)


adriansuhov:
Tested by me & all is good now

PR #506, opened 2018-12-07
Add VmOfToday utility

lizzha:


lizzha:
@ader1990 updated

Utilities/VmOfToday.ps1
```diff
@@ -0,0 +1,207 @@
...
+				$result = [ordered]@{}
+				$result.Add("Resource Group Name", $rg.ResourceGroupName)
+				$result.Add("Role Name", $vmData.roleName)
+				$result.Add("Public IP Address", $vmData.PublicIP)
+				$result.Add("SSH Port", $vmData.SSHPort)
+				$result.Add("User Name", $xmlSecrets.secrets.linuxTestUsername)
+				$result.Add("Password", $xmlSecrets.secrets.linuxTestPassword)
+				$result.Add("Instance Size", $vmData.InstanceSize)
+				$result.Add("Location", $vmData.Location)
+				$result.Add("VM Creation Time", $rg.Tags.CreationTime)
```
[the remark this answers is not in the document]
lizzha: move the creation time to beginning, locks are already listed

Utilities/VmOfToday.ps1
```diff
@@ -0,0 +1,242 @@
+##############################################################################################
+# VmOfToday.ps1
```
juhlee-microsoft: License should be at the first line.

Utilities/VmOfToday.ps1
```diff
@@ -0,0 +1,242 @@
...
+
+.INPUTS
+
+
+.NOTES
+	Creation Date:
+	Purpose/Change:
+
+.EXAMPLE
+
```
juhlee-microsoft: Can you put "how to use" with parameters here?

Utilities/VmOfToday.ps1
```diff
@@ -0,0 +1,242 @@
...
+.PARAMETER
+	-Operation, Release or GetAndLock
+	-UserName, the user that want to get and lock or release the resource group
+	-AzureSecretsFile, the path of Azure secrets file
+	-RgPattern, the pattern of resource group name
+	-TagName, the tag name to filter resource group
+	-TagValue, the tag value to filter resource group
+
+.INPUTS
```
juhlee-microsoft: It is nice to add example of parameters here and show any use case

Utilities/VmOfToday.ps1
```diff
@@ -0,0 +1,242 @@
...
+	}
+	if (-not $vmFound) {
+		Write-LogErr "Cannot find valid running VM."
+	}
+	if ($usersToNotify.Count -gt 0) {
+		"USERS_TO_NOTIFY=$($usersToNotify -join ',')" | Out-File env.properties -Encoding ASCII
+	}
+}
+
+Function Unlock-LockedResourceGroup($UserName, $AzureSecretsFile, $RgPattern, $TagName, $TagValue){
```
juhlee-microsoft: I am not sure if this is right, because this can unlock the locked RG forcefully. Sometimes I should keep RG for long time investigation. If this can unlock the locked RG, my work is not safe. What is the purpose of this function? @lizzha

lizzha: This script is to get and lock latest RG with specific name pattern and tag value, and also release the locked RGs. The lock it will add or remove is named by the user who triggered the Jenkins job, or named by the timer.
So it will not touch other RGs or locks.

lizzha:
@juhlee-microsoft  "If this is in LISAv2, then anyone can remove Azure RG with Secret file. I suggest this should be under another authentication before execution."
Could you help me understand this comment? Anyone has a secret file or has admin access to the subscription would be able to remove an RG or a lock on RG, it's not any issue related with the script.

Utilities/VmOfToday.ps1
juhlee-microsoft: This script has **overriding power** for RG management. If used in aggressive way or abusive, it could be unlocking RG which should be locked. And I still don't understand why we need this utility script.

juhlee-microsoft:
> @juhlee-microsoft "If this is in LISAv2, then anyone can remove Azure RG with Secret file. I suggest this should be under another authentication before execution."
> Could you help me understand this comment? Anyone has a secret file or has admin access to the subscription would be able to remove an RG or a lock on RG, it's not any issue related with the script.

If anyone know our subscription and secret file, any accident unlocks the locked RG. It will remove RG eventually. Again, this script is good one, but I am not sure if we keep it in public.

Utilities/VmOfToday.ps1
lizzha: The script will only unlock the locks named "Timer Lock" or $UserName, on the resource groups with specified pattern and tag value. So if it would unlock a resource group, it would be run on purpose rather than by accident. 
In the case it's run on purpose, without this script, user still can run the Remove-AzureRmResourceLock command directly to unlock a resource group. Again it's not a problem of the script, but rather subscription/user management.

The utility is for always keeping the latest Linux-next VMs in Azure, with the compiled source code, so that developers can use the VM and quickly apply and test their patch.

lizzha:
@juhlee-microsoft  As I commented above, RGs would hard to be unlocked by accident using this script, but rather unlocked on purpose.

Issue #493, opened 2018-12-05
Common performance reporting framework in lisav2


ader1990:
To compare the results for the performance testing, we need a framework that:

- is generic - should work for all performance results
- has machine readable output

For genericity, all performance test cases should have a JSON output, with the name 

> TestCaseName_perf_results.json

The JSON file should contain an array of objects, with the object having the following structure
{
  "Id": ValueID,
 "ResultType1": Value1,
 "ResultType2": Value2,
 "ResultType3": Value3
}

Is mandatory to have at least the Id field.

This has to be implemented for all performance test cases to be able to have a comparison of the results.
If the test does not implement, there will be no result comparison possible.

PR: https://github.com/LIS/LISAv2/pull/491 is a proof of concept for this implementation.

The changes in the other test cases should require refactoring and adding this feature.
This feature is backwards compatible with the previous workflow.

PR #508, opened 2018-12-07
perf_ntttcp: perf_stor: implement perf result comparison framework

ader1990:


ader1990:
This patch fixes issue: https://github.com/LIS/LISAv2/issues/493

The iperf3/fio and ntttcp perf tests were updated according to the issue description and resolution.

PR #510, opened 2018-12-09
 Add case to build upstream kernel from git source

lizzha:


lizzha:
@ader1990 it's not duplicated. We need the VM that contains upstream source as well as compilation history, for developers to use. With the VM, they can apply patch and quickly build the kernel. The -customKernel parameter only supports install kernel from package.

adriansuhov:
@lizzha So why not just modify customKernelInstall.sh for your needs and use BVT-VERIFY-DEPLOYMENT-PROVISION? This way you'll have a VM with a custom kernel installed

lizzha:
@adriansuhov that's a good suggestion, let me update customKernelInstall.sh

ader1990:
@lizzha please check again the functionality here: https://github.com/LIS/LISAv2/blob/master/Testscripts/Linux/customKernelInstall.sh#L289

You can reuse that functionality or change it if need be, but we wouldn't want to end up with duplicate kernel build code in lisav2.

and please do not remove the labels without a proper discussion on why they have been added.

PR #519, opened 2018-12-11
Fix installing linux upstream kernel with -CustomKernel

lizzha:
Tested on Ubuntu 16.04, Ubuntu 18.04 and CentOS 7.5

### DIFF
--- a/Libraries/Azure.psm1
+++ b/Libraries/Azure.psm1
@@ -1935,6 +1935,7 @@ Function Deploy-ResourceGroups ($xmlConfig, $setupType, $Distro, $getLogsIfFaile
 Function Check-SSHPortsEnabled($AllVMDataObject) {
     Write-LogInfo "Trying to Connect to deployed VM(s)"
     $timeout = 0
+    $retryCount = 20
     do {
         $WaitingForConnect = 0
         foreach ( $vm in $AllVMDataObject) {
@@ -1958,7 +1959,7 @@ Function Check-SSHPortsEnabled($AllVMDataObject) {
         if ($WaitingForConnect -gt 0) {
             $timeout = $timeout + 1
             Write-LogInfo "$WaitingForConnect VM(s) still awaiting to open port $port .."
-            Write-LogInfo "Retry $timeout/20"
+            Write-LogInfo "Retry $timeout/$retryCount"
             sleep 3
             $retValue = "False"
         } else {
@@ -1966,7 +1967,7 @@ Function Check-SSHPortsEnabled($AllVMDataObject) {
             $retValue = "True"
         }
 
-    } While (($timeout -lt 20) -and ($WaitingForConnect -gt 0))
+    } While (($timeout -lt $retryCount) -and ($WaitingForConnect -gt 0))
 
 	if ($retValue -eq "False") {
 		foreach ($vm in $AllVMDataObject) {
--- a/Libraries/Azure.psm1
+++ b/Libraries/Azure.psm1
@@ -1958,7 +1958,7 @@ Function Check-SSHPortsEnabled($AllVMDataObject) {
         if ($WaitingForConnect -gt 0) {
             $timeout = $timeout + 1
             Write-LogInfo "$WaitingForConnect VM(s) still awaiting to open port $port .."
-            Write-LogInfo "Retry $timeout/100"
+            Write-LogInfo "Retry $timeout/20"
             sleep 3
             $retValue = "False"
         } else {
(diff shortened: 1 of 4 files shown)


Libraries/Azure.psm1
```diff
@@ -1958,7 +1958,7 @@ Function Check-SSHPortsEnabled($AllVMDataObject) {
         if ($WaitingForConnect -gt 0) {
             $timeout = $timeout + 1
             Write-LogInfo "$WaitingForConnect VM(s) still awaiting to open port $port .."
-            Write-LogInfo "Retry $timeout/100"
+            Write-LogInfo "Retry $timeout/20"
             sleep 3
```
ader1990: add a variable maxTimeout = 20 and use it when needed

Libraries/CommonFunctions.psm1
```diff
@@ -194,7 +194,7 @@ function Install-CustomKernel ($CustomKernel, $allVMData, [switch]$RestartAfterU
 			$CustomKernelLabel = $CustomKernel
 			foreach ( $vmData in $allVMData )
 			{
-				Copy-RemoteFiles -uploadTo $vmData.PublicIP -port $vmData.SSHPort -files ".\Testscripts\Linux\$scriptName,.\Testscripts\Linux\DetectLinuxDistro.sh" -username $user -password $password -upload
+				Copy-RemoteFiles -uploadTo $vmData.PublicIP -port $vmData.SSHPort -files ".\Testscripts\Linux\$scriptName,.\Testscripts\Linux\utils.sh" -username $user -password $password -upload
```
ader1990: are you sure detectlinuxdistro is not used anymore?

ader1990:
@lizzha what is the time of such a build on Ubuntu and Rhel?
usually, the kernel build can take a very long time if not properly optimized.

please check the docs here and make sure that the kernel build is optimized as much as possible: 
https://github.com/LIS/lis-pipeline/blob/master/docs/Configure-RhelKernelBuilder.md
https://github.com/LIS/lis-pipeline/blob/master/docs/Configure-UbuntuKernelBuilder.md

lizzha:
@ader1990 on a D4_v3 VM, with make -j4, on Ubuntu the total kernel build and install time was about 35 minutes, while on CentOS 7.5 the time was about 55 minutes.
I tried to change the make thread to 8, which is 2 times of the core count, the build time doesn't get reduced, even a bit longer.

So I updated the script to set the make thread number same as the core number, and also use ccache as the docs suggest.

lizzha:
The logic for other branches are not changed, so I tested the following -CustomKernel value for sanity check
- a local rpm package
- a public URL points to a deb package

ader1990:
> @ader1990 on a D4_v3 VM, with make -j4, on Ubuntu the total kernel build and install time was about 35 minutes, while on CentOS 7.5 the time was about 55 minutes.
> I tried to change the make thread to 8, which is 2 times of the core count, the build time doesn't get reduced, even a bit longer.
> 
> So I updated the script to set the make thread number same as the core number, and also use ccache as the docs suggest.

the ccache won't help on the first build, unfortunately. the only thing that helps is the core count (but only if you have an excess of 16 or 32 cores). what helps a lot is the parallel compression (in the docs).
you can use a different vm size (minimum of 16 cores and premium storage) plus the pigz hack -- I think that can get you to 15mins-20mins.

ader1990:
I have tested with multiple deb kernel files at once on hyperv and it works as expected.
@adriansuhov please make sure this change does not affect the azure kernel functionality.
@lizzha let me know when the ccache is removed (it does not bring any value on cold build) and when you consider it as final patch (you can use the no merge label until this is ready).

lizzha:
@ader1990 thanks.
I would keep the ccache, as it speeds up recompilation, which will be needed for developers.  I'm doing some test with parallel compression, will let you know when the PR is ready for merge.

lizzha:
@ader1990 I tested the parallel compression, but it doesn't show improvements in built time. So I would keep the patch as-is. 
Could you please approve and merge?

ader1990:
> @ader1990 I tested the parallel compression, but it doesn't show improvements in built time. So I would keep the patch as-is.
> Could you please approve and merge?

sure, we can revisit this build kernel thing when we do the general refactor, so that we do not need to keep two separate code bases of kernel build -- one in lisav2 and one in lis-pipeline.

ader1990:
> @ader1990 I tested the parallel compression, but it doesn't show improvements in built time. So I would keep the patch as-is.
> Could you please approve and merge?

I am waiting @adriansuhov sign off, to make sure the azure kernel pipelines are in order after merge, will merge asap after his sign off, thanks.

PR #535, opened 2018-12-14
Update name of setupType for one VM

Tong-Xu:
The following is results.
[LISAv2 Test Results Summary]
Test Run On           : 12/14/2018 10:29:10
ARM Image Under Test  : canonical : ubuntuserver : 18.04-lts : Latest
Total Time (dd:hh:mm) : 0:0:12

    1 BVT-SWAP-CHECK                                                     PASS                 7.91


LiliDeng:
@lizzha @simonxiaoss Let know about your thoughts for keeping stress type.

lizzha:
@ader1990 the setup type should be classified by VM settings, rather than by test scenarios. For the convenience of knowing OverrideVmSize for stress cases, I would suggest to put all the stress cases together in a same xml.

ader1990:
@LiliDeng I still don't agree with removing the Stress setup type, as it makes it harder to understand the test case definition. But if you want to still want to make this change, as it is not breaking any functionality,  I will approve it.

PR #536, opened 2018-12-14
CAPTURE-VHD-BEFORE-TEST.ps1: Added HISTSIZE=0

chvalean:
Official documentation recommends to clear the shell history.
https://docs.microsoft.com/en-us/azure/virtual-machines/linux/create-upload-generic

This is useful for custom modified vhd images that might not be properly cleaned up.


Testscripts/Windows/CAPTURE-VHD-BEFORE-TEST.ps1
```diff
@@ -9,7 +9,6 @@ function Main {
     try {
         $testResult = $null
         $captureVMData = $allVMData
-        # region CONFIGURE VM FOR N SERIES GPU TEST
```
LiliDeng: remove # endregion also

LiliDeng:
there were some existed typo if you want to fix
# so we need to detect the distro **vefore** deprovisioning
Write-LogInfo "Could not **potentialy** fix Ubuntu Bionic waagent. Continue execution..."

chvalean:
Thank you @LiliDeng for the feedback, I re-submitted the PR in #538

Issue #481, opened 2018-12-03
Test tag consolidation request


juhlee-microsoft:
As of Dec-13, 2018, there are so many tags in TestCase xml files. Need to consolidate them and minimize the tag counts.

===== Tag Details =====

Name                           Value
----                           -----
password                       2
backup                         10
hv_storvsc                     7
network                        77
fcopy                          8
restart                        1
shutdown                       2
lsvmbus                        1
heartbeat                      2
vm_paused                      1
rdma                           20
nested-kvm                     12
cd                             2
hv_module                      5
provision                      6
linux_setting                  5
swap                           1
syscall                        1
secureboot                     4
migration                      10
stress                         8
kdump                          6
disk                           4
timesync                       8
dpdk                           8
sriov                          42
floppy_disk                    1
memory                         27
prod_checkpoint                9
smb                            1
nested-hyperv                  7
storage                        49
fstab                          1
time                           3
boot                           2
pci_hyperv                     4
wala                           9
cpu                            2
nested                         19
ethtool                        7
ltp                            1
hv_daemon                      1
kvp                            12
nmi                            2
gpu                            4
hv_netvsc                      24
mtab                           1
vmbus                          4

chvalean:
Some suggestions:
disk and fstab might go under storage
mtab and smb and swap under storage
vm_paused should go away
floppy_disk and cd should be generalized under a misc section

chvalean:
> vm_paused can not go away, but has to find alternative. I recommend 'power_management' including restart, shutdown, & boot.
> I can not put them in 'misc' because 'misc' does not inform the user to raise the confidence of information. Let's leave them as they are until we find better one.

a shorter name would be vm_states, and indeed it can cover those other mentioned tags

juhlee-microsoft:
Close because we only hold P0

PR #539, opened 2018-12-14
VERIFY-IFUP-IFDOW-ETH0-RMMOD-MODPROBE-HV_NETVSC-Synthetic: Renamed to…

chvalean:
… VERIFY-IFUP-IFDOWN-RELOAD-HV_NETVSC-Synthetic

Name is too long.
Also align area and tag for this test case to match the other network tests

### DIFF
--- a/XML/TestCases/FunctionalTests-Network.xml
+++ b/XML/TestCases/FunctionalTests-Network.xml
@@ -444,7 +444,7 @@
         <Priority>2</Priority>
     </test>
     <test>
-        <testName>VERIFY-IFUP-IFDOW-ETH0-RMMOD-MODPROBE-HV_NETVSC-Synthetic</testName>
+        <testName>VERIFY-IFUP-IFDOWN-RELOAD-HV_NETVSC-Synthetic</testName>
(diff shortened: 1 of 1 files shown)


XML/TestCases/FunctionalTests-Network.xml
```diff
@@ -457,8 +457,8 @@
         <files>.\Testscripts\Linux\verify-ifdown-ifup-eth0-rmmod-modprobe-hv-netvsc.sh,.\TestScripts\Linux\utils.sh</files>
         <Platform>Azure</Platform>
         <Category>Functional</Category>
-        <Area>Synthetic</Area>
-        <Tags>hv_netvsc</Tags>
+        <Area>NETWORK</Area>
+        <Tags>network</Tags>
```
LiliDeng: Maybe hv_netvsc is more proper?

chvalean: will probably consolidate the tags when https://github.com/LIS/LISAv2/issues/481 will be handled.

PR #541, opened 2018-12-17
Add display powerstate of VM for Get-AzureVMs.ps1

LiliDeng:
Make it as optional parameter, since it will cost 2 mins to get all VM's status.
By default, run this PS file cost within 1 min.
Fix an existed exception.
Multiple ambiguous overloads found for "op_Subtraction" and the argument count: "2".
At E:\lili\master\bugfix\LISAv2\Utilities\Get-AzureVMs.ps1:148 char:17
+ ...             $ageInDays = ([datetime]::Now - $rg.Tags.CreationTime).Da ...
+                 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : NotSpecified: (:) [], MethodException
    + FullyQualifiedErrorId : MethodCountCouldNotFindBest

Usage .\Get-AzureVMs.ps1 -IncludeState


Utilities/Get-AzureVMs.ps1
```diff
@@ -95,10 +80,11 @@ function Get-VMAgeFromDisk()
 #Get all VMs and enumerate thru them adding items to results list.
 $allVMs = Get-AzureRmVM
 $allRGs = Get-AzureRmResourceGroup
-
+if ($IncludeState) {
```
lizzha: if $IncludeState is set, it actually get the VMs twice, the first time is not necessary
$allVMs = Get-AzureRmVM
$allVMsWithState = Get-AzureRmVM -Status

Better to change to
if ($IncludeState) {
    $allVMs = Get-AzureRmVM -Status
} else {
    $allVMs = Get-AzureRmVM
}

LiliDeng: @lizzha Thanks for review, seems these two commands return different objects.

PR #408, opened 2018-11-20
Fixing BVT-CAPTURE-VHD-BEFORE-TEST deprovision issue

asofron:

PR #543, opened 2018-12-18
Fix issue that Provisioning.UseCloudInit is not set to n, and other two typos

LiliDeng:
Fix issue for not execute disable Provisioning.UseCloudInit
Fix other 2 typo issues


lizzha:
It's an increment fix for https://github.com/LIS/LISAv2/pull/408

PR #544, opened 2018-12-18
This is for fix when testname contains lower case

LiliDeng:
This is for fix when testname contains lower case, it won't collect this case info
Some pipelines still use previous names like PERF-NETWORK-TCP-SINGLE-CONNECTION-THROUGHPUT-Synthetic
Let know if no need update here but update test cases name in pipeline


lizzha:
yes, better to update case name in pipeline

LiliDeng:
Ok, close it. Thanks.

PR #548, opened 2018-12-18
lisav2: speed up kernel boot fail check

ader1990:


ader1990:
This is a patch that makes the lisav2 fail faster if there is a kernel panic.

ader1990:
Before this patch, lisav2 failed in approximately 14 minutes.
```
ARM Image Under Test  : Canonical : UbuntuServer : 18.04-DAILY-LTS : latest

Total Test Cases      : 1 (0 Pass, 0 Fail, 1 Abort)

Total Time (dd:hh:mm) : 0:0:13

    1 BVT-CORE-VERIFY-LIS-MODULES                                     ABORTED                 13.9 

```

ader1990:
After the patch, lisav2 should fail in less than 8 minutes:
```

[azure_boot] Test Run On           : 12/18/2018 17:38:20
[azure_boot] ARM Image Under Test  : Canonical : UbuntuServer : 18.04-DAILY-LTS : latest
[azure_boot] Total Test Cases      : 1 (0 Pass, 0 Fail, 1 Abort)
[azure_boot] Total Time (dd:hh:mm) : 0:0:7
[azure_boot] XML File              : TestConfiguration
[azure_boot] 
[azure_boot]    ID TestCaseName                                                 TestResult TestDuration(in minutes) 
[azure_boot] ------------------------------------------------------------------------------------------------------
[azure_boot]     1 BVT-CORE-VERIFY-LIS-MODULES                                     ABORTED                 7.58 
[azure_boot] 
[azure_boot] 
[azure_boot] Logs can be found at C:\LISAv2\MW72\TestResults\2018-18-12-17-38-16-2369\BVT-CORE-VERIFY-LIS-MODULES
[azure_boot]
```

Libraries/HyperV.psm1
```diff
@@ -701,7 +700,7 @@ Function Restart-AllHyperVDeployments($allVMData)
     {
         Start-HyperVGroupVMs -HyperVGroupName $VM.HyperVGroupName -HyperVHost $VM.HyperVHost
     }
-	$isSSHOpened = Check-SSHPortsEnabled -AllVMDataObject $AllVMData
+	$isSSHOpened = Is-VmAlive -AllVMDataObject $AllVMData
```
LiliDeng: $isVmAlive =?

Libraries/Azure.psm1
```diff
@@ -2487,3 +2422,36 @@ Function Add-DefaultTagsToResourceGroup {
         Write-LogErr "EXCEPTION in Add-DefaultTagsToResourceGroup() : $ErrorMessage at line: $ErrorLine"
     }
 }
+
+function Get-AzureBootDiagnostics {
+    param($vm)
```
LiliDeng: for this function, it should named like Check-KernelPanic?

Libraries/LISAV2-Framework.psm1
```diff
@@ -1398,4 +1398,47 @@ Function Do-TestCleanUp($CurrentTestResult, $testName, $DeployedServices, $Resou
...
+            } else {
+                $port = $vm.SSHPort
+            }
+
+            $out = Test-TCP -testIP $vm.PublicIP -testport $port
+            if ($out -ne "True") {
+                Write-LogInfo "Connecting to $($vm.PublicIP) and port $port failed."
+                $deadVms += 1
+                # Note(v-advlad): Check for kernel panic once every 3 retries on Azure
+                if (($retryCount % 3 -eq 0) -and ($TestPlatform -eq "Azure") -and (Get-AzureBootDiagnostics $vm)) {
```
lizzha: will it increase the check time for non-kernelpanic cases?

Libraries/HyperV.psm1
ader1990: will update

Libraries/Azure.psm1
ader1990: I was thinking too, to change the name of this method.
I think Check-AzureVmKernelPanic is a good name, as it is only available on Azure

Libraries/LISAV2-Framework.psm1
ader1990: not really, as the diagnostics are retrieved very fast and only after the 3rd retry.
Usually, at the 3rd retry, the VM should be already up and running.

Libraries/Azure.psm1
ader1990: @LiliDeng I have split the method in two, one that gets the boot diagnostics and one that checks for kernel panic

Libraries/LISAV2-Framework.psm1
ader1990: and it will happen only on azure with linux vm type, where it speeds up the execution by more than 2x on a single test case if the test case has a kernel boot issue

Libraries/HyperV.psm1
ader1990: fixed

Utilities/VmOfToday.ps1
```diff
@@ -134,8 +134,8 @@ Function Lock-LatestResourceGroup($UserName, $AzureSecretsFile, $RgPattern, $Tag
 		try {
 			Write-LogInfo "Start to validate the deployments in resource group $($rg.ResourceGroupName)"
 			$vmData = Get-AllDeploymentData -ResourceGroups $rg.ResourceGroupName
-			$isSSHOpened = Check-SSHPortsEnabled -AllVMDataObject $vmData
-			if (!$isSSHOpened) {
```
ader1990: @lizzha should I fix also this code? 
```
$isSSHOpened = Check-SSHPortsEnabled -AllVMDataObject $vmData
			if (!$isSSHOpened) {
```

Check-SSHPortsEnabled always returned "False" or "True" as strings, and this check was wrong, as it always passes.
Was there a reason behind this logic?

Libraries/LISAV2-Framework.psm1
```diff
@@ -1398,4 +1398,59 @@ Function Do-TestCleanUp($CurrentTestResult, $testName, $DeployedServices, $Resou
...
+            }
+        }
+
+        if ($deadVms -gt 0) {
+            Write-LogInfo "$deadVms VM(s) still waiting to open port $port."
+            Write-LogInfo "Retrying $retryCount/$MaxRetryCount in 3 seconds."
+            Start-Sleep -Seconds 3
+            $retValue = "False"
+        } else {
+            Write-LogInfo "ALL VMs' ports are open now."
```
asofron: "All VM`s ports are open now."

Utilities/VmOfToday.ps1
lizzha: @ader1990 please fix this, thanks

PR #552, opened 2018-12-19
lisav2: allow multiple test categories, comma separated

ader1990:
This commit allows specifying multiple test categories, instead of none or only one.


ader1990:
This commit allows specifying multiple test categories, instead of none or only one.

PR #568, opened 2018-12-31
Refactor LISAv2

lizzha:


ader1990:
Can you please move all the logic in powershell modules so that we can write unit tests in pester for all the core framework?

Basically, Run-Lisav2 should become a shim layer that call the main function from a module.

juhlee-microsoft:
I am not sure if I can do code-review 186 files change. Can you also do document update in the same PR? It looks many changes.

ader1990:
> @ader1990 could you describe more specifically which logic to move?

all of it, run-lisav2.ps1 should import one PS module and call the init main function like this Start-LisaV2 &param

ader1990:
This way, we can use pester to test all the framework methods and leave almost nothing in the core framework to be tested as a full run.

lizzha:
@ader1990 I don't think it's necessary to move all the logic in Run-LISAv2.ps1 to a module. If we do so, then Run-LISAv2.ps1 is not needed, and the PS module actually is another kind of Run-LISAv2.
But I agree that we can put as much logic as possible from Run-LISAv2.ps1 to modules, but Run-LISAv2.ps1 still should contain the necessary steps of calling functions.

TestControllers/TestController.psm1
```diff
@@ -0,0 +1,421 @@
...
+		}
+
+		$timeout = 300
+		if ($CurrentTestData.Timeout) {
+			$timeout = $CurrentTestData.Timeout
+		}
+
+		# Run test script
+		if ($CurrentTestData.TestScript) {
+			$testResult = Run-TestScript -CurrentTestData $CurrentTestData `
```
mbivolan: If any powershell script executed in this method throws an error, the testing will stop as the only catch is in Run-LisaV2.ps1.
Try to wrap this command in a try/catch, and is case an error is thrown, set the test on FAIL and continue with the next ones.

ader1990:
@lizzha please create a list of to do's you consider fit for your PR (what you want to achieve and what has been achieved in the current state).

ader1990:
> @ader1990 I don't think it's necessary to move all the logic in Run-LISAv2.ps1 to a module. If we do so, then Run-LISAv2.ps1 is not needed, and the PS module actually is another kind of Run-LISAv2.
> But I agree that we can put as much logic as possible from Run-LISAv2.ps1 to modules, but Run-LISAv2.ps1 still should contain the necessary steps of calling functions.

If you do not move all the logic in modules, you will not be able to unit test the code.
Moreover, maybe one day we would want to put lisav2 on PowerShell galery, in order to install it with Install-Module command from Windows. Then, we require to have it in a module only.

TestControllers/TestController.psm1
lizzha: fixed

santoshx:
01/08/2019 07:04:11 : [ERROR] Failed to update configuration files. '-XMLSecretFile [FilePath]' is not provided.

Test continues even without a secret file.
I suppose this is alright?

lizzha:
@santoshx yes, if xml secret file is not provided, the test can go on if the powershell session is already authenticated to Azure

santoshx:
01/08/2019 07:43:00 : [INFO ] Checking if Hyper-V VM group 'LISAv2-OneVM12Disks-sx-FX40-636825499804' exists on localhost...
01/08/2019 07:43:01 : [WARN ] Hyper-V VM group LISAv2-OneVM12Disks-sx-FX40-636825499804 does not exist

Framework checks and WARNs when there isn't a pre-existing group?
Should we have necessary groups already created before starting the test?

lizzha:
>           01/08/2019 07:43:00 : [INFO ] Checking if Hyper-V VM group 'LISAv2-OneVM12Disks-sx-FX40-636825499804' exists on localhost...
> 01/08/2019 07:43:01 : [WARN ] Hyper-V VM group LISAv2-OneVM12Disks-sx-FX40-636825499804 does not exist
> Framework checks and WARNs when there isn't a pre-existing group?
> Should we have necessary groups already created before starting the test?

No, this behavior is the same as current behavior, the step is to make sure the group doesn't exist

santoshx:
> > ```
> >       01/08/2019 07:43:00 : [INFO ] Checking if Hyper-V VM group 'LISAv2-OneVM12Disks-sx-FX40-636825499804' exists on localhost...
> > ```
> > 01/08/2019 07:43:01 : [WARN ] Hyper-V VM group LISAv2-OneVM12Disks-sx-FX40-636825499804 does not exist
> > Framework checks and WARNs when there isn't a pre-existing group?
> > Should we have necessary groups already created before starting the test?
> 
> No, this behavior is the same as current behavior, the step is to make sure the group doesn't exist

Then I think there is no use to WARN when group doesn't exist. 
We may WARN if there is already a group and framework going to use it.

lizzha:
@santoshx Oh I misunderstood this message, just double checked the message is in the clean up stage. In method Delete-HyperVGroup, it warns that the group to delete doesn't exist.

lizzha:
@santoshx  to be accurate, the message is written in function Delete-HyperVGroup, but Delete-HyperVGroup is called when deploying VMs to make sure no group with the same name exists, if it exists, the group would be deleted.
Feel free to file a PR to improve it.

Run-LisaV2.ps1
```diff
@@ -78,11 +81,9 @@ Param(
 	[string] $CustomParameters = "",
 	[string] $OverrideVMSize = "",
 	[switch] $EnableAcceleratedNetworking,
-	[string] $OverrideHyperVDiskMode = "",
 	[switch] $ForceDeleteResources,
 	[switch] $UseManagedDisks,
 	[switch] $DoNotDeleteVMs,
-	[switch] $DeployVMPerEachTest,
```
ader1990: why are you removing DeployVMPerEachTest ?

Run-LisaV2.ps1
```diff
@@ -29,6 +29,9 @@
 
 #>
 ###############################################################################################
+using Module ".\TestControllers\TestController.psm1"
```
ader1990: you don t need to import test controller, you have to do import test controller  only in hyperv and azure test controller

Run-LisaV2.ps1
lizzha: The design is changed, DeployVMPerEachTest is not used anymore. Cases will share deployments, if they have same setupType, OverrideVMSize, networking, diskType, osDiskType, switchName.

Run-LisaV2.ps1
lizzha: will remove

Testscripts/Windows/SETUP-NET-Add-NIC.ps1
```diff
@@ -26,7 +26,7 @@
 
 param(
     [String] $TestParams,
-    [String] $VMName = $AllVMData.RoleName
```
asofron: $VMName variable in needed in script .Or use $AllVMData.RoleName instead of  $VMName

Testscripts/Windows/SETUP-NET-Switch-NIC.ps1
```diff
@@ -6,10 +6,7 @@
  Switch an existing NIC (with a certain MAC address) to a different network.
 #>
 
-param(
-    [String] $VMName=$AllVMData.RoleName,
-    [String] $TestParams
-)
+param([string] $TestParams, [object] $AllVMData)
```
asofron: $AllVMData.RoleName ?

Testscripts/Windows/SETUP-RemoveIsoFromDvd.ps1
```diff
@@ -11,7 +11,7 @@
     to null instead.
 #>
 
-param([String] $TestParams)
+param([string] $TestParams, [object] $AllVMData)
```
asofron: $AllVMData.RoleName ?

Testscripts/Windows/SETUP-SR-IOV-Configure.ps1
```diff
@@ -13,7 +13,7 @@
     3. Set up SR-IOV on VM2
     Optional: Set up an internal network on VM2
 #>
-param ([string] $TestParams)
+param([string] $TestParams, [object] $AllVMData)
```
asofron: $AllVMData.RoleName ?

Run-LisaV2.ps1
lizzha: added back

ader1990:
@lizzha I have tested the patch with the developer comparison pipeline and it works perfectly. will sign off, thanks.

ader1990:
@lizzha where can we find the new documentation on the changes and on the global variables?

Libraries/IntegrationServiceLibrary.psm1
```diff
@@ -0,0 +1,1098 @@
...
+	Creation Date:
+	Purpose/Change:
+
+.EXAMPLE
+
+
+#>
+###############################################################################################
+
+function Wait-VMHeartbeatOK {
```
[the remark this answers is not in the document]
lizzha: updated

lizzha:
@ader1990  for the LISAv2 new design, I will write a document and upload to sharepoint. For the global variables, I will file a new PR to update the document

ader1990:
@lizzha great, thanks.

Libraries/IntegrationServiceLibrary.psm1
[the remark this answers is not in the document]
ader1990: I still see in this file a lot of hyper-v related methods that can be moved to the hyper-v module.
@lizzha do you plan to leave them in this module? We can move them in another PR if you prefer to merge this asap. let me know if you plan to move them in a separate PR so that I can approve this PR.

PR #570, opened 2019-01-04
BVT-NET-ETHTOOL-CHANGE-RINGBUFFER: add 4 more retries

ader1990:


Libraries/CommonFunctions.psm1
```diff
@@ -896,7 +896,7 @@ Function Test-TCP($testIP, $testport)
 	return $isConnected
 }
 
-Function Copy-RemoteFiles($uploadTo, $downloadFrom, $downloadTo, $port, $files, $username, $password, [switch]$upload, [switch]$download, [switch]$usePrivateKey, [switch]$doNotCompress, $maxRetry=20) #Removed XML config
+Function Copy-RemoteFiles($uploadTo, $downloadFrom, $downloadTo, $port, $files, $username, $password, [switch]$upload, [switch]$download, [switch]$usePrivateKey, [switch]$doNotCompress, $maxRetry=20, $maxRetryDownload=50, $maxRetryUpload=10) #Removed XML config
```
lizzha: Is there any reason to set download retry to 50, and upload retry to 10?
I have done an uncommitted change to refactor this Copy-RemoteFiles, just want to understand whether we can reduce the retry number.
@iamshital could you answer this question?

lizzha: if the two retry counts can be set to the same, it's not needed to add two more parameters

ader1990: what I did is to preserve the backwards compatibility. this patch is possible to not be needed, as the ethtool issue seems to not reproduce all the time using plink.exe

LiliDeng: You can see the log from upstream job, this case cost more than 5 hours to download files.
I already repro this issue with size A2, maybe this PR is not needed.

iamshital: > 
> 
> Is there any reason to set download retry to 50, and upload retry to 10?
> I have done an uncommitted change to refactor this Copy-RemoteFiles, just want to understand whether we can reduce the retry number.
> @iamshital could you answer this question?

Because, in case of failures, most of the times, we observed download was prone to errors than upload. (May be due to slow network / large number of download files). Hence download attempts were more than upload.

PR #572, opened 2019-01-07
Fix Xfstesting-cifs and KDUMP test case issues

LiliDeng:
Fix cases of TestCategory Community Area Xfstesting-cifs
Fix cases of TestCategory Functional Area KDUMP after disable root


ader1990:
@LiliDeng can you please put in the commit title the problem you fixed and a small commit description?
"Fix test case issues" is very confusing.

LiliDeng:
@ader1990 
For KDUMP, fix issues for hardcode root.
For Xfstesting-cifs, test case will abort since test script can't find the config file "xfstests-config.config"

juhlee-microsoft:
@ader1990 please sign off if you are ok.

PR #594, opened 2019-01-11
Enable test against existing VMs in existing resource group

lizzha:


ader1990:
@lizzha we already have a parameter called RGIdentifier that can be used for this reason.
why not add a flag [switch]$ReuseExistentRG to Run-Lisav2 that can be used to use the RGIdentifier exactly as it is? Otherwise we will have a useless parameter RgIndentifier at the same time that is used for kind of exactly the same thing.

Also, this should be implemented for HyperV, so that no hyperv groups should be created.

ader1990:
@lizzha By the way, this is a very good feature and it will speed up around 1 minute, maybe more, each Azure test.

lizzha:
@ader1990 good suggestion, I will update the code. 
Please note that the change is only for running test against existing VMs in existing resource group.
It's to do item to enable test against an empty resource group.

iamshital:
Hi Liz, This is good. I would recommend to to keep -ExistingRG as string parameter.

There are two requirements -
1. **Add support to deploy VMs in empty resource group.**
    Follow up : Add support to cleanup all the resources also except resource group.

2. **Add support to deploy VMs in resource group where AvailibilitySet is already created.**
    Follow up : Add support to cleanup all the resources except AvailibilitySet

lizzha:
@iamshital I would take the advice to use $RGIdentifier with a switch $UseExistingRG, because $RGIdentifier is mandatory parameter, if we introduce a string $ExistingRG parameter, it's duplicated information with $RGIdentifier.
What's the scenario of deploying VMs in an existing RG with AvailibilitySet created?

iamshital:
> @iamshital I would take the advice to use $RGIdentifier with a switch $UseExistingRG, because $RGIdentifier is mandatory parameter, if we introduce a string $ExistingRG parameter, it's duplicated information with $RGIdentifier.
> What's the scenario of deploying VMs in an existing RG with AvailibilitySet created?

There are scenarios in Azure Host Build Testing, where Azure compute team gives us the resource group with AvailibilitySet or Resource Group **pinned** to some test cluster. In this case, If we get the AvailibilitySet pinned to some cluster, then we want to deploy the VMs in using that availibility set only.

lizzha:
The PR is updated, it enables
1. Test against existing VMs in a resoure group or HyperV group
2. Test against empty resource group/HyperV group
After test running, by default it will delete the resources in the group, the group itself will be kept. 
Use -DoNotDeleteVMs if you don't want to delete the resources

@iamshital for the scenario of existing resource group with AvailibilitySet, it can be done later as it would be lower priority

LiliDeng:
Hi Liz, how to specify use which existed group using RGIdentifier?

lizzha:
>           Hi Liz, how to specify use which existed group using RGIdentifier?

-RGIdentifer <exactresourcegroup> -UseExistingRG

PR #605, opened 2019-01-15
Refactor function copy-remotefiles

lizzha:


LiliDeng:
Besides split upload and download into different methods and decrease the max retry. any other changes?

lizzha:
>           Besides split upload and download into different methods and decrease the max retry. any other changes?

remove duplicated logic in compress and not compress

PR #611, opened 2019-01-16
Fixing broken function after refactor  in FCOPY/RUNTIME_MEMORY/DYNAMIC_MEMORY

asofron:
Fixed failing functions in RUNTIME_MEMORY Tests
Fixed failing functions in FCOPY tests
Fixed failing functions in DYNAMIC_MEMORY Tests

Validation Tests : 

FCOPY:

VHD Under Test        : Ubuntu_18.04.1.vhdx
Test Category         : Functional
Test Area             : FCOPY
Total Test Cases      : 8 (8 Pass, 0 Fail, 0 Abort)
Total Time (dd:hh:mm) : 0:0:32

    1 FCOPY-BASIC                                                                       PASS                 2.28 
    2 FCOPY-FILE-EXISTS                                                                 PASS                 2.07 
    3 FCOPY-OVERWRITE                                                                   PASS                 2.08 
    4 FCOPY-LARGE-FILE                                                                  PASS                 8.11 
    5 FCOPY-NON-ASCII                                                                   PASS                 2.44 
    6 FCOPY-REPEATED-DELETE                                                             PASS                 3.31 
    7 FCOPY-DISABLE-ENABLE                                                              PASS                 7.84 
    8 FCOPY-NEGATIVE                                                                    PASS                 1.84


Libraries/CommonFunctions.psm1
```diff
@@ -1670,7 +1670,7 @@ Function Is-StressNgInstalled {
     Copy-RemoteFiles -uploadTo $VMIpv4 -port $VMSSHPort -files $FILE_NAME -username $user -password $password -upload
     # execute command
     $retVal = Run-LinuxCmd -username $user -password $password -ip $VMIpv4 -port $VMSSHPort `
-        -command "echo $password | sudo -S cd /root && chmod u+x ${FILE_NAME} && sed -i 's/\r//g' ${FILE_NAME} && ./${FILE_NAME}"
+        -command "echo $password | cd /home/$user && chmod u+x ${FILE_NAME} && sed -i 's/\r//g' ${FILE_NAME} && ./${FILE_NAME}"
```
juhlee-microsoft: here all your changes are from /root to /home/$user. if that requires 'root' priv, how can we handle it here?

lizzha: do we need to add  -runAsSudo ?

juhlee-microsoft: Basically $user should be non-root because ofInstalled. if this is installed by root, then it is nice to have -runAdSudo. If you are not sure, then leave is as you have until it break. We can fix it later.

asofron: Added -runAsSudo retesting to check if it breaks anything

PR #612, opened 2019-01-16
LIS RPM: Add test to cross verify build version with source version

DivyaJyothiKasam:
Build version is retrieved from modinfo and source version
is retrieved from lis-next hv_compat.h

### DIFF
--- a/Testscripts/Linux/LIS-VERSION-CHECK.sh
+++ b/Testscripts/Linux/LIS-VERSION-CHECK.sh
@@ -1,0 +1,87 @@
+#!/bin/bash
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the Apache License.
+
+##################################################################
+#
+# Description:
+# This script compares LIS source version (#define HV_DRV_VERSION)
+# with version retrieved from command modinfo
+#
+##################################################################
+
+function check_lis_version
+{
+	version=$1
+	sourceversion=$2
+	LogMsg "Detected modinfo version is $version"
+	LogMsg "Version found in source code is $sourceversion"
+	if [ "$version" = "$sourceversion" ]
+	then
+		LogMsg "Detected version and Source version are same"
+	else
+		LogMsg "Detected version and Source version are different"
+		updateSummary "LIS version check is failed"
+		SetTestStateFailed
+		exit 0
+	fi
+}
+
+function check_lis_version_hex()
+{
+	version=$1
+	sourceversion_hex=$2
+	LogMsg "Detected modinfo version is $version"
+	LogMsg "Version found in source code hex value :$sourceversion_hex"
+	version1=$(echo "$version"|tr -d .)
+	version_hex=$(echo "obase=16;ibase=10; $version1" | bc |tr -d '" "')
+	if [ "$version_hex" = "$sourceversion_hex" ]
+	then
+		LogMsg "Detected version and Source version are same for hex value"
+	else
+		LogMsg "Detected version and Source version are different for hex value"
+		updateSummary "LIS hex version check failed"
+		SetTestStateFailed
+		exit 0
+	fi
+}
+
+# Source utils.sh
+. utils.sh || {
+	echo "ERROR: unable to source utils.sh!"
+	echo "TestAborted" > state.txt
+	exit 2
+}
+
+# Source constants file and initialize most common variables
+UtilsInit
+# Check wget
+which wget > /dev/null 2>&1
+if [ $? -ne 0 ]; then
+    install_package wget
+    check_exit_status "wget install" "exit"
+fi
+
+# To check LIS rpms are installed or not
+rpm -qa | grep "kmod-microsoft-hyper-v" && rpm -qa | grep "microsoft-hyper-v"
+if [ $? -ne 0 ]
+then
+    LogMsg "No LIS RPM's are detected. Skipping test."
+    SetTestStateAborted
+    exit 0
+else
+    version=$(modinfo "hv_vmbus" | grep version: | head -1 | awk '{print $2}')
+    for i in 5 6 7
+    do
+        rm -rf hv_compat.h
+        wget https://raw.githubusercontent.com/LIS/lis-next/"$version"/hv-rhel$i.x/hv/include/linux/hv_compat.h
+        check_exit_status "Download file hv_compat.h" "exit"
+        sourceversion=$(grep 'define HV_DRV_VERSION' hv_compat.h|cut -d '"' -f 2)
+        sourceversion_hex=$(grep 'define _HV_DRV_VERSION' hv_compat.h|cut -d ' ' -f 3|tr -d '0x')
+        check_lis_version $version $sourceversion
+        check_lis_version_hex $version $sourceversion_hex
+    done
+fi
+
+UpdateSummary "LIS version check is success"
+SetTestStateCompleted
--- a/XML/TestCases/BVT.xml
+++ b/XML/TestCases/BVT.xml
@@ -1,5 +1,16 @@
 <TestCases>
     <test>
+        <testName>BVT-LIS-DRIVER-VERSION-CHECK</testName>
+        <testScript>LIS-VERSION-CHECK.sh</testScript>
+        <files>.\TestScripts\Linux\LIS-VERSION-CHECK.sh,.\TestScripts\Linux\utils.sh</files>
+        <setupType>OneVM</setupType>
+        <Platform>Azure</Platform>
+        <Category>BVT</Category>
+        <Area>BVT</Area>
+        <Tags>lis_provision</Tags>
+        <Priority>0</Priority>
+    </test>
+    <test>
         <testName>BVT-IS-ROOT-PASSWORD-DELETED</testName>
         <testScript>BVT-IS-ROOT-PASSWORD-DELETED.py</testScript>
         <files>.\Testscripts\Linux\BVT-IS-ROOT-PASSWORD-DELETED.py,.\Testscripts\Linux\azuremodules.py</files>


DivyaJyothiKasam:
[LISAv2 Test Results Summary]
Test Run On           : 01/16/2019 10:55:57
VHD Under Test        : Centos_7.5_x64.vhdx
Total Time (dd:hh:mm) : 0:0:5

    1 BVT-VERSION-CHECK                                                                 PASS                 3.22

XML/TestCases/BVT.xml
```diff
@@ -1,4 +1,15 @@
 <TestCases>
+    <test>
+        <testName>BVT-LIS-DRIVER-VERSION-CHECK</testName>
+        <testScript>LIS-VERSION-CHECK.sh</testScript>
+        <files>.\TestScripts\Linux\LIS-VERSION-CHECK.sh,.\TestScripts\Linux\utils.sh</files>
+        <setupType>OneVM</setupType>
+        <Platform>HyperV</Platform>
+        <Category>BVT</Category>
+        <Area>BVT</Area>
+        <Tags>provision</Tags>
```
chvalean: tag should be lis_provision, or only lis, given that it targets specifically only that.

Testscripts/Linux/LIS-VERSION-CHECK.sh
```diff
@@ -0,0 +1,78 @@
...
+}
+# Source constants file and initialize most common variables
+UtilsInit
+# Check wget
+wget -V > /dev/null 2>&1
+if [ $? -ne 0 ]; then
+    yum install wget -y
+    check_exit_status "wget install" "exit"
+fi
+version=$(modinfo "hv_vmbus" | grep version: | head -1 | awk '{print $2}')
```
chvalean: this entire test must be conditioned by a check to see if LIS 4.x is installed - rpm's are present.
Otherwise this test is NA, will exit with TestSkipped in the future when we implement that exit status.

Testscripts/Linux/LIS-VERSION-CHECK.sh
```diff
@@ -0,0 +1,78 @@
+#!/bin/bash
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the Apache License.
```
chvalean: Please include the description of the test and what it does/check in a header at the start of the script.

XML/TestCases/BVT.xml
```diff
@@ -1,4 +1,15 @@
 <TestCases>
+    <test>
+        <testName>BVT-LIS-DRIVER-VERSION-CHECK</testName>
+        <testScript>LIS-VERSION-CHECK.sh</testScript>
+        <files>.\TestScripts\Linux\LIS-VERSION-CHECK.sh,.\TestScripts\Linux\utils.sh</files>
+        <setupType>OneVM</setupType>
+        <Platform>HyperV</Platform>
```
juhlee-microsoft: Only Hyper-V?

Testscripts/Linux/LIS-VERSION-CHECK.sh
```diff
@@ -0,0 +1,78 @@
...
+# Source constants file and initialize most common variables
+UtilsInit
+# Check wget
+wget -V > /dev/null 2>&1
+if [ $? -ne 0 ]; then
+    yum install wget -y
+    check_exit_status "wget install" "exit"
+fi
+version=$(modinfo "hv_vmbus" | grep version: | head -1 | awk '{print $2}')
+rm -rf hv_compat.h
```
juhlee-microsoft: Line number 56 to 78 are duplicated 3 times except the path. Put them in a single for loop.

Testscripts/Linux/LIS-VERSION-CHECK.sh
```diff
@@ -0,0 +1,78 @@
...
+wget https://raw.githubusercontent.com/LIS/lis-next/"$version"/hv-rhel7.x/hv/include/linux/hv_compat.h
```
juhlee-microsoft: IS it only for RHEL? what about other distro?

XML/TestCases/BVT.xml
chvalean: this is suffice to run on only one platform, the check relates to the LIS version so it is enough to run either on Azure or only on Hyper-V, no need to cover both platforms in this particular case.

Testscripts/Linux/LIS-VERSION-CHECK.sh
chvalean: Agree, the entire test needs to be conditioned to:
- RHEL family - including centos and Oracle.
- OS family 6, 7 (6.x and 7.x releases)
- condition the above to hv-rhel6.x or hv-rhel7.x, depending on above.
- LIS RPM installed/presence - already mentioned in review.

DivyaJyothiKasam: > Agree, the entire test needs to be conditioned to:
> 
> * RHEL family - including centos and Oracle.
@chvalean will skip test for non RHEL family
> * OS family 6, 7 (6.x and 7.x releases)
@chvalean Do you want to skip this test for OS family 5.x .. Why?

Testscripts/Linux/LIS-VERSION-CHECK.sh
```diff
@@ -0,0 +1,78 @@
+#!/bin/bash
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the Apache License.
+function CheckVersion()
```
ader1990: replace the method names with a bash like format camel_case
also, make sure the name of the methods are clear, in your case you should have something like:

check_lis_version

Testscripts/Linux/LIS-VERSION-CHECK.sh
```diff
@@ -0,0 +1,78 @@
...
+		UpdateSummary "LIS version check is success"
+		SetTestStateCompleted
+	else
+		LogMsg "Detected version and Source version are different"
+		updateSummary "LIS version check is failed"
+		SetTestStateFailed
+		exit 0
+	fi
+}
+function CheckHexVersion()
```
ader1990: ditto:
check_hex_lis_version

Testscripts/Linux/LIS-VERSION-CHECK.sh
```diff
@@ -0,0 +1,78 @@
...
+# Source utils.sh
+. utils.sh || {
+	echo "ERROR: unable to source utils.sh!"
+	echo "TestAborted" > state.txt
+	exit 2
+}
+# Source constants file and initialize most common variables
+UtilsInit
+# Check wget
+wget -V > /dev/null 2>&1
```
ader1990: use which to check for commands

Testscripts/Linux/LIS-VERSION-CHECK.sh
```diff
@@ -0,0 +1,78 @@
...
+rm -rf hv_compat.h
+wget https://raw.githubusercontent.com/LIS/lis-next/"$version"/hv-rhel7.x/hv/include/linux/hv_compat.h
+check_exit_status "Download file hv_compat.h" "exit"
+sourceversion=$(grep 'define HV_DRV_VERSION' hv_compat.h|cut -d '"' -f 2)
+sourceversion_hex=$(grep 'define _HV_DRV_VERSION' hv_compat.h|cut -d ' ' -f 3|tr -d '0x')
+CheckVersion $version $sourceversion
+CheckHexVersion $version $sourceversion_hex
+
+rm -rf hv_compat.h
+wget https://raw.githubusercontent.com/LIS/lis-next/"$version"/hv-rhel6.x/hv/include/linux/hv_compat.h
```
ader1990: create a method and reuse the code

Testscripts/Linux/LIS-VERSION-CHECK.sh
chvalean: Absolutely if you want to cover all of them.
However on Azure I don't think you have RedHat 5.x

juhlee-microsoft:
@DivyaJyothiKasam  any update for comment?

XML/TestCases/BVT.xml
DivyaJyothiKasam: Done

Testscripts/Linux/LIS-VERSION-CHECK.sh
DivyaJyothiKasam: put a condition, whether LIS rpm's are installed or not

XML/TestCases/BVT.xml
DivyaJyothiKasam: Done

Testscripts/Linux/LIS-VERSION-CHECK.sh
DivyaJyothiKasam: Removed the duplicated code

Testscripts/Linux/LIS-VERSION-CHECK.sh
DivyaJyothiKasam: The test don't care about distro flavor. It only checks whether LIS rpms are installed or not. If installed it moves ahead does the version check.
I don't see need for distro check.

Testscripts/Linux/LIS-VERSION-CHECK.sh
DivyaJyothiKasam: done

Testscripts/Linux/LIS-VERSION-CHECK.sh
DivyaJyothiKasam: Done

DivyaJyothiKasam:
> @DivyaJyothiKasam any update for comment?

sorry for delay. Am on leave.

Testscripts/Linux/LIS-VERSION-CHECK.sh
DivyaJyothiKasam: done

DivyaJyothiKasam:
[LISAv2 Test Results Summary]
Test Run On : 01/23/2019 13:36:39
ARM Image Under Test : OpenLogic : CentOS : 7.5 : 7.5.20180815
Total Test Cases : 1 (1 Pass, 0 Fail, 0 Abort)
Total Time (dd:hh:mm) : 0:0:4

    1 BVT-LIS-DRIVER-VERSION-CHECK                                                              PASS                 1.65

Testscripts/Linux/LIS-VERSION-CHECK.sh
```diff
@@ -0,0 +1,87 @@
...
+	echo "TestAborted" > state.txt
+	exit 2
+}
+
+# Source constants file and initialize most common variables
+UtilsInit
+# Check wget
+which wget > /dev/null 2>&1
+if [ $? -ne 0 ]; then
+    yum install wget -y
```
ader1990: here you should use install_packages method that works on all distros

Testscripts/Linux/LIS-VERSION-CHECK.sh
DivyaJyothiKasam: done

Testscripts/Linux/LIS-VERSION-CHECK.sh
```diff
@@ -0,0 +1,87 @@
+#!/bin/bash
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the Apache License.
+
+##################################################################
+#
+# Description:
+# This script compares LIS source version (#define HV_DRV_VERSION)
+# with version retrived from command modinfo
```
LiliDeng: retrieved

LiliDeng: make sure this file indent consistent

Testscripts/Linux/LIS-VERSION-CHECK.sh
```diff
@@ -0,0 +1,87 @@
...
+	LogMsg "Detected modinfo version is $version"
+	LogMsg "Version found in source code hex value :$sourceversion_hex"
+	version1=$(echo "$version"|tr -d .)
+	version_hex=$(echo "obase=16;ibase=10; $version1" | bc |tr -d '" "')
+	if [ "$version_hex" = "$sourceversion_hex" ]
+	then
+		LogMsg "Detected version and Source version are same for hex value"
+	else
+		LogMsg "Detected version and Source version are different for hex value"
+		updateSummary "LIS hex version check is failed"
```
LiliDeng: is can be removed

juhlee-microsoft:
@DivyaJyothiKasam  please update for the rest of open comment.

Testscripts/Linux/LIS-VERSION-CHECK.sh
DivyaJyothiKasam: Done

Testscripts/Linux/LIS-VERSION-CHECK.sh
DivyaJyothiKasam: Done

Testscripts/Linux/LIS-VERSION-CHECK.sh
DivyaJyothiKasam: Done.

DivyaJyothiKasam:
[LISAv2 Test Results Summary]
Test Run On           : 01/25/2019 07:14:32
ARM Image Under Test  : OpenLogic : CentOS : 7.5 : 7.5.20180815
Total Time (dd:hh:mm) : 0:0:22

    1 BVT-LIS-DRIVER-VERSION-CHECK                                                      PASS                 4.98

chvalean:
> There is case BVT-CORE-VERIFY-LIS-MODULES-VERSION, is it possible to combine the logic together, I saw it also use modinfo to get the version.

That test is a generic - all distros applicable one.
This one instead is rather specific to only when LIS RPMs are present, so the method and login are different.

Testscripts/Linux/LIS-VERSION-CHECK.sh
```diff
@@ -0,0 +1,87 @@
+#!/bin/bash
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the Apache License.
+
+##################################################################
+#
+# Description:
+# This script compares LIS source version (#define HV_DRV_VERSION)
+# with version retrieved from command modinfo
+#
```
chvalean: Please also mention that this test case is applicable only when using the latest released LIS version set of RPMs.
Otherwise maybe also include a check that this might fail for the simple reason of a previous LIS, there would be no match between them, but that would be a false-positive failure.

chvalean:
Added another comment, please also test this with a previous LIS version installed.
This test will only work in ideal conditions, otherwise it's prone to multiple points of failure, more or less false-positive ones.

Testscripts/Linux/LIS-VERSION-CHECK.sh
DivyaJyothiKasam: This test case is applicable all LIS releases not just latest.

DivyaJyothiKasam:
> Added another comment, please also test this with a previous LIS version installed.
> This test will only work in ideal conditions, otherwise it's prone to multiple points of failure, more or less false-positive ones.

Please state where it can fail.

Testscripts/Linux/LIS-VERSION-CHECK.sh
chvalean: You're right, I missed the part where you download lis-next/"$version"

LiliDeng:
this case only can run against distro which has rpm command, for other distro, it will abort

PR #618, opened 2019-01-17
Fix test  abort when python3 installed + no softlink for it.

LiliDeng:
Run a case against image which can invoke python directly - pass
Run a case against image which install python3, but no softlink for it.


Libraries/CommonFunctions.psm1
```diff
@@ -336,6 +336,8 @@ function Run-TestScript {
         Write-LogInfo "${scriptLoc} -TestParams $scriptParameters -AllVmData $VmData -TestProvider $TestProvider -CurrentTestData $CurrentTestData"
         $testResult = & "${scriptLoc}" -TestParams $scriptParameters -AllVmData $VmData -TestProvider $TestProvider -CurrentTestData $CurrentTestData
     } elseif ($scriptExtension -eq "py") {
+        Run-LinuxCmd -Username $Username -password $Password -ip $VMData.PublicIP -Port $VMData.SSHPort `
+             -Command "which python || (which python3 && ln -s /usr/bin/python3 /usr/bin/python && which python)" -runAsSudo
```
lizzha: the last "which python" is not necessary?

LiliDeng: I just want to make sure in any condition, it will return the output for which python.
But I don't see the output from the log, is any thing change for Run-LinuxCmd?

lizzha: No, there's no change in Run-LinuxCmd
If you remove the last "which python", it would still output the python path, which would be the path of python3

LiliDeng: Ok

ader1990:
@LiliDeng In the upcoming RHEL8 there will be no python binary, only python3 or python2. is this scenario handled by the current code? Please check https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/8-beta/html/8.0_beta_release_notes/new-features#web_servers_databases_dynamic_languages_2

LiliDeng:
I will add logic for this situation, thanks @ader1990

LiliDeng:
Run a case against image which can invoke python directly - pass
Run a case against image which install python3, but no softlink for it - pass
Run a case against CoreOS - pass

LiliDeng:
I miss some places which has its own ps1 file.

Libraries/CommonFunctions.psm1
```diff
@@ -322,6 +322,16 @@ function Run-TestScript {
             Write-LogInfo "Constants file uploaded to: $($VM.RoleName)"
         }
     }
+    if($CurrentTestData.files -imatch ".py") {
+        $pythonPath = Run-LinuxCmd -Username $Username -password $Password -ip $VMData.PublicIP -Port $VMData.SSHPort `
+            -Command "which python || which python2 || which python3" -runAsSudo
+        if (($pythonPath -imatch "python2") -or ($pythonPath -imatch "python3")) {
+            $python_Path = $pythonPath.Substring(0, $pythonPath.LastIndexOf("/") + 1)
```
ader1990: replace python_Path with pythonPathSymlink to avoid confusing var names

LiliDeng: Thanks, fix and test - pass

PR #623, opened 2019-01-18
Replace deprecated ifconfig with iproute2 - ip commands

chvalean:
ifconfig use is deprecated and newer distributions no longer include the package by default.
Replace ifconfig/ifdown/ifup commands with the new set of ip commands.

Rewrite VLAN interface create to not use interface config file, but rather only to current boot session.
*This model should be carried to further code refactor for configuring the secondary interfaces as well.*

Also rename testcase name and file "verify-ifdown-ifup-eth0-rmmod-modprobe-hv-netvsc.sh" to a shorter version.

---

Azure:
ARM Image Under Test  : Canonical : UbuntuServer : 18.04-LTS : latest
BVT-CORE-RELOAD-MODULES-SMP                                                       PASS                11.36 <- also pass on Hyper-V
VERIFY-RESTART-INTERFACE-RELOAD-NETVSC                                            PASS                 2.05
HyperV:
NET-MAX-SYNTHETIC-NICS                                                            PASS                 3.72
NET-VLAN-TAGGING                                                                  PASS                 7.32
NET-VLAN-TRUNKING                                                                 PASS                 6.72

### DIFF
--- a/Libraries/CommonFunctions.psm1
+++ b/Libraries/CommonFunctions.psm1
@@ -851,30 +851,24 @@ function Enable-SRIOVInAllVMs($allVMData, $TestProvider)
 				$vmCount += 1
 				if ($sriovOutput -imatch "DATAPATH_SWITCHED_TO_VF")
 				{
-					$AfterIfConfigStatus = $null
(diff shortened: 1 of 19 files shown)


chvalean:
Testing the changes now, will request review after posting the results.

PR #627, opened 2019-01-21
hyperv: don't apply checkpoints on the first run and retry to remove snaspshots

ader1990:
hyperv: retry to remove snapshots on failure

hyperv: don't revert unneeded snasphots when the first test case is run, as it is not needed

Execution tested on HyperV:

```
    1 BVT-CORE-VERIFY-LIS-MODULES                                                       PASS                 0.95
```

### DIFF
--- a/TestProviders/TestProvider.psm1
+++ b/TestProviders/TestProvider.psm1
@@ -35,7 +35,7 @@ Class TestProvider
 		return $null
 	}
 
-	[void] RunSetup($VmData, $CurrentTestData, $TestParameters) {}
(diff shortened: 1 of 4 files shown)


TestProviders/TestProvider.psm1
```diff
@@ -35,7 +35,7 @@ Class TestProvider
 		return $null
 	}
 
-	[void] RunSetup($VmData, $CurrentTestData, $TestParameters) {}
+	[void] RunSetup($VmData, $CurrentTestData, $TestParameters, $ApplyCheckPoint) {}
```
lizzha: the interface in AzureProvider.psm1 should also be changed? 
To make it general, suggest to name the parameter $FirstRun or similar?

PR #628, opened 2019-01-21
Fix test case issues for BVT category

LiliDeng:
1 ClearOS the boot config path is different
2 no wget tool in oracle distro by default
3 add entry for more distro in timeclock cases
4 Password policy check before it run without sudo, currently framework using sudo, case will fail, update code according, and add one more situation when password contains part of username.


Testscripts/Linux/BVT-NET-IFUP-IFDOWN.sh
```diff
@@ -9,6 +9,9 @@ LoopCount=10
 
 PingCheck()
 {
+    if [[ $(detect_linux_distribution) == oracle ]]; then
+        yum install -y wget
```
lizzha: no need to detect the distro, we cannot assume other distros has wget preinstalled.
just run: install_package wget

lizzha:
also please post the test results

Testscripts/Linux/BVT-NET-IFUP-IFDOWN.sh
LiliDeng: make sense, will update and post results later, thanks for review

LiliDeng:
image

PR #643, opened 2019-01-23
azure: added HBC and HCS family support

adriansuhov:
HBS and HCS VM families were considered as H family until now and that was causing various issues in lisav2 testing.


chvalean:
@juhlee-microsoft  for ack.
We'll create a new issue to rewrite the function handling the VM sizes, current implementation of loop conditions for each VM size is impossible to maintain and has to be streamlined.

PR #644, opened 2019-01-23
dpdk: enable netvsc driver and small improvements

ader1990:


Testscripts/Linux/dpdkSetup.sh
```diff
@@ -118,12 +118,18 @@ function install_dpdk () {
...
+	LogMsg "CONFIG_RTE_LIBRTE_VDEV_NETVSC_PMD flag enable on ${1}"
+	ssh "${1}" "sed -i 's/^CONFIG_RTE_LIBRTE_VDEV_NETVSC_PMD=n/CONFIG_RTE_LIBRTE_VDEV_NETVSC_PMD=y/g' $HOMEDIR/$dpdkSrcDir/config/common_base"
+	check_exit_status "${1} CONFIG_RTE_LIBRTE_VDEV_NETVSC_PMD=y"
+
 	ssh "${1}" "cd $HOMEDIR/$dpdkSrcDir && make config O=$DPDK_BUILD T=$DPDK_BUILD"
 	LogMsg "Starting DPDK build make on ${1}"
-	ssh "${1}" "cd $HOMEDIR/$dpdkSrcDir/$DPDK_BUILD && make -j8 && make install"
+	proc_nr=$(cat /proc/cpuinfo | grep vendor | wc -l)
+	ssh "${1}" "cd $HOMEDIR/$dpdkSrcDir/$DPDK_BUILD && make -j${proc_nr} && make install"
 	check_exit_status "dpdk build on ${1}"
```
chvalean: suggest to do a check_exit_status after make and another one after make install?
Then it can be clear to report if the compilation or the install failed.

ader1990: make install is just a shim call that copies some files to another folder, I don t think it is relevant to check each calls

Testscripts/Linux/dpdkSetup.sh
```diff
@@ -118,12 +118,18 @@ function install_dpdk () {
 	else
 		LogMsg "dpdk build with default DST IP ADDR on ${1}"
 	fi	
-	LogMsg "MLX_PMD flag enabling on ${1}"
-	ssh "${1}" "sed -i 's/^CONFIG_RTE_LIBRTE_MLX4_PMD=n/CONFIG_RTE_LIBRTE_MLX4_PMD=y/g' $HOMEDIR/$dpdkSrcDir/config/common_base"
-	check_exit_status "${1} CONFIG_RTE_LIBRTE_MLX4_PMD=y"
+	LogMsg "CONFIG_RTE_LIBRTE_MLX._PMD enable ${1}"
+	ssh "${1}" "sed -ri 's,(MLX._PMD=)n,\1y,' $HOMEDIR/$dpdkSrcDir/config/common_base"
```
ader1990: @justin-parus @juhlee-microsoft these config changes are okay with you?
before the change, only mlx4_pmd was enabled, but I have seen in the azure dpdk docs that mlx5 and netvsc are also required.
Do you have any insight on this?

Thank you.

justin-parus: I have started using dpdkSetupAndRunTest.sh in combination with dpdkUtils.sh as it provides a more complete set of tools to test dpdk. dpdkUtils.sh provides `Install_Dpdk()` which installs dpdk with support for both mlx 4,5 and follows the method from the official [doc](https://docs.microsoft.com/en-us/azure/virtual-network/setup-dpdk).
```
ssh ${install_ip} "cd ${LIS_HOME}/${DPDK_DIR} && make config T=x86_64-native-linuxapp-gcc"
ssh ${install_ip} "sed -ri 's,(MLX._PMD=)n,\1y,' ${LIS_HOME}/${DPDK_DIR}/build/.config"
```

The files dpdkSetupAndRunTest.sh + dpdkUtils.sh + DPDK_TESTCASE_DRIVER.ps1 acts as a "micro"-framework within the LISAv2 in that it provides some boilerplate code for dpdk tests. dpdk-testPMD.sh, dpdk-testFWD.sh, and dpdk-testFailsafe.sh all use this setup. 

dpdkSetup.sh and dpdkTestPmd.sh were great starting points and much of the code in the newer dpdk files was derived from them. I think it would be okay to deprecate these older files. Check with @juhlee-microsoft to be sure they aren't being used elsewhere or aren't piping data to a PowerBI graph. Last I heard this was already moved to use dpdk-TestPMD.sh instead

justin-parus: Further dpdkSetupAndRunTest.sh + dpdkUtils.sh + DPDK_TESTCASE_DRIVER.ps1 are in a functional state right now, but still could use much improvement and are open to suggestion. For example, what @chvalean mention above about checking the status of the build is something that has been needed for awhile now. As I have seen builds mysteriously fail (my guess is something with `make -j`) before and waste pipeline time.

ader1990: @justin-parus dpdkSetup.sh is still used by the test cases mentioned by @juhlee-microsoft 

To give more context, these are the changes required for building OVS/VPP/NFF-GO and ANS with DPDK support.
This is work in progress in order to make a dpdk setup which works for the new test cases planned for DPDK.

ader1990: @justin-parus My main concern is that these changes might affect the other test cases, though it should not be the case, which is why I wanted your insight on this matter.

juhlee-microsoft:
@ader1990 I know this is low priority now. Can you complete this PR or pull back? It's been 28 days.

ader1990:
> @ader1990 I know this is low priority now. Can you complete this PR or pull back? It's been 28 days.

closing this for the moment, will sync with Stephen H. and make sure this is not required.

PR #660, opened 2019-01-28
Storage tests: fix when disk is not partitioned

LiliDeng:
In pipeline the command output is 
image

Results from pipeline
image

It threw below exception
01/28/2019 07:46:23 : [INFO ] .\tools\plink.exe -t -pw ****** -P 22 lisa@52.191.132.54 "echo ****** | sudo -S /sbin/fdisk -l | grep /dev/sd"
01/28/2019 07:46:29 : [INFO ] /sbin/fdisk -l | grep /dev/sd executed successfully in 5.33 seconds.
01/28/2019 07:46:29 : [ERROR] Cannot convert value "a" to type "System.Int64". Error: "Input string was not in a correct format." at line: 39

After fix:
image


Testscripts/Windows/DiskHotAddRemove-Parallel.ps1
```diff
@@ -38,7 +38,7 @@ function Main {
         Write-LogInfo "Verifying if data disk is added to the VM: Running fdisk on remote VM"
         $fdiskOutput = Run-LinuxCmd -username $user -password $password -ip $AllVMData.PublicIP -port $AllVMData.SSHPort -command "/sbin/fdisk -l | grep /dev/sd" -runAsSudo
         foreach($line in ($fdiskOutput.Split([Environment]::NewLine))) {
-            if($line -imatch "Disk /dev/sd[^ab]" -and [int64]($line.Split()[4]) -eq (([int64]($diskSizeinGB) * [int64]1073741824))){
+            if($line -imatch "Disk /dev/sd[^ab]:" -and [int64]($line.Split()[4]) -eq (([int64]($diskSizeinGB) * [int64]1073741824))){
```
lizzha: why did it cause the test abort? 
It looks like the output of fdisk -l varies on different distros, if so, have you tested on different distros?

LiliDeng: Below output is from the VM when specify -CustomKernel 'proposed-azure'
Disk /dev/sdc doesn't contain a valid partition table
Disk /dev/sdb: 7516 MB, 7516192768 bytes
/dev/sdb1             128    14678015     7338944    7  HPFS/NTFS/exFAT
Disk /dev/sda: 32.2 GB, 32212254720 bytes
/dev/sda1   *        2048    62914559    31456256   83  Linux
Disk /dev/sdc: 32.2 GB, 32212254720 bytes

This line 'Disk /dev/sdc doesn't contain a valid partition table' is not expected, it caused case abort when convert a string into int. Add colon will exclude this line.

I will run it against other distro.

lizzha: I see, then it should be OK. But it's good to test on other distros.

LiliDeng: Thanks much for review, I will post results once done.

PR #661, opened 2019-01-29
Support NVMe disk perf test without formatting disk

xian123:


LiliDeng:
@xian123 Please run functional testing and regression testing, make sure not break current testing, and new testing runs well. Thanks!

chvalean:
let's close this and default to PR #663 instead

adriansuhov:
Hi! PR https://github.com/LIS/LISAv2/pull/663 has a few additional things that were discussed in the past days:
 - Dynamic test matrix - for each LSv2 vm size, the table will alter itself automatically. So not only the numJob param will be dynamically set (as is here), but also startQdepth and maxQdepth.
Full details are in PR description

PR #672, opened 2019-01-30
Add support for OL

vyadavmsft:

PR #673, opened 2019-01-31
Fix issue bring by https://github.com/LIS/LISAv2/pull/672

LiliDeng:
Test results will be posts after a few minutes.


LiliDeng:
@juhlee-microsoft Vivek is fixing shell check and PS check issues.

image

PR #683, opened 2019-01-31
hyperv: reuse failed test case vm

### LINKED ISSUE #599: hyperv: bring back snapshot feature
Before PR #568 , if a test case failed on HyperV, a snapshot was created with the failure and for the following test the VM was reverted to its original pre test state.

We need to bring back that feature.

ader1990:
The refactor removed snapshot usage on Hyper-V.

This patch brings back the snapshot feature usage in Hyper-V.

On HyperV platform, the VMs are reused for each test case that asks for the same VM setup type in order to decrease resource consumption.

General workflow (changes of this patch are in bold):

1. create vm
2. create checkpoint icabase for that vm
3. first test case is run on the vm
4. **if the first test case fails, create a checkpoint for the failed state.**
5. **apply icabase checkpoint**
6. second test case is run on the same vm, etc.

Fixes #599

### DIFF
--- a/TestControllers/TestController.psm1
+++ b/TestControllers/TestController.psm1
@@ -286,12 +286,12 @@ Class TestController
 					-GlobalConfig $this.GlobalConfig -AllVMData $VmData
 			}
 
+			# Run setup script if any
+			$this.TestProvider.RunSetup($VmData, $CurrentTestData, $testParameters, $ApplyCheckpoint)
+
 			if (!$this.IsWindows) {
 				GetAndCheck-KernelLogs -allDeployedVMs $VmData -status "Initial" | Out-Null
 			}
-
-			# Run setup script if any
-			$this.TestProvider.RunSetup($VmData, $CurrentTestData, $testParameters, $ApplyCheckpoint)
 
 			# Upload test files to VMs
 			if ($CurrentTestData.files) {
@@ -376,8 +376,9 @@ Class TestController
 					if ( $this.TestIterations -ne 1 ) {
 						$case.testName = "$($originalTestName)-$testIterationCount"
 					}
+					Write-LogInfo "$($case.testName) started running."
 					$executionCount += 1
-					if (!$vmData -or $this.DeployVMPerEachTest) {
+					if (!$vmData -or $this.DeployVMPerEachTest -or $this.ForceDeleteResources) {
 						# Deploy the VM for the setup
 						$vmData = $this.TestProvider.DeployVMs($this.GlobalConfig, $this.SetupTypeTable[$setupType], $this.SetupTypeToTestCases[$key][0], `
 							$this.TestLocation, $this.RGIdentifier, $this.UseExistingRG)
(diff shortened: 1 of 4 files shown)


TestControllers/TestController.psm1
```diff
@@ -397,8 +397,9 @@ Class TestController
 					if ($lastResult.TestResult -ne "PASS") {
 						if ($this.ForceDeleteResources) {
 							$this.TestProvider.DeleteTestVMS($vmData, $this.SetupTypeTable[$setupType], $this.UseExistingRG)
+						} else {
+							$this.TestProvider.CreateFailedCheckpoint($vmData, $case.testName, $lastResult.TestResult)
 						}
-						$vmData = $null
```
lizzha: removing this line will affect test on azure, the design is to keep the failed VM on azure, and deploy new VMs for subsequent test

TestControllers/TestController.psm1
```diff
@@ -397,8 +397,9 @@ Class TestController
 					if ($lastResult.TestResult -ne "PASS") {
 						if ($this.ForceDeleteResources) {
 							$this.TestProvider.DeleteTestVMS($vmData, $this.SetupTypeTable[$setupType], $this.UseExistingRG)
+						} else {
+							$this.TestProvider.CreateFailedCheckpoint($vmData, $case.testName, $lastResult.TestResult)
```
lizzha: if you check the method RunTestCaseCleanup of HyperVProvider, it already creates checkpoint for failed cases.
I suggest not to add another method CreateFailedCheckpoint to the TestProvider.
Instead, I think we can add a property to TestProvider: $KeepFailedVM, for AzureProvider, it's $true, for HyperVProvider, it's $false. And if it's $true, set the $vmData to $null, so that new VMs will by deployed

lizzha: The property can be added to TestController

ader1990: will be done, thanks for the review. currently, I need to test this change on all platforms and see what are all the implications.

ader1990: > RunTestCaseCleanup

@lizzha you said "if you check the method RunTestCaseCleanup of HyperVProvider, it already creates checkpoint for failed cases." -- which is not correct, the code for some reason does not do that. I will investigate further and have a fix for that.

TestControllers/TestController.psm1
ader1990: I added a new provider property for this: ReuseVmOnFailure

TestProviders/HyperVProvider.psm1
```diff
@@ -58,37 +60,32 @@ Class HyperVProvider : TestProvider
...
 					return $null
 				}
 
-				# Create the initial checkpoint
-				Create-HyperVCheckpoint -VMData $AllVMData -CheckpointName "ICAbase"
-				$allVMData = Check-IP -VMData $AllVMData
-			}
-			else
-			{
+				Inject-HostnamesInHyperVVMs -allVMData $allVMData
```
lizzha: should the VM be restarted after injecting hostname?

asofron:
VMs do not  remain stuck anymore and checkpoint is created .

PR #687, opened 2019-02-01
Add utility script to create service principal

lizzha:


Utilities/CreateServicePrincipal.ps1
```diff
@@ -0,0 +1,247 @@
...
+    Write-Prompt "2) Full privileges, includes Contributor"
+    Write-Prompt "Please choose by entering 1 or 2:"
+    $Privilege = Read-Host
+
+    Write-Host "Assign roles to Service Principal..."
+    if ($Privilege -eq 2) {
+        New-AzureRmRoleAssignment -RoleDefinitionName "Contributor" -ApplicationId $ClientId
+    }
+    else {
+        New-AzureRmRoleAssignment -RoleDefinitionName "Virtual Machine Contributor" -ApplicationId $ClientId
```
iamshital: Why not "Owner" for full privileges?

lizzha: isn't Contributor enough for LISAv2 test?

Utilities/CreateServicePrincipal.ps1
```diff
@@ -0,0 +1,246 @@
...
+        the string specified with the parameter FirstChar
+    .OUTPUTS
+        [String]
+    .NOTES
+        Written by Simon Wåhlin, blog.simonw.se
+        I take no responsibility for any issues caused by this script.
+    .FUNCTIONALITY
+        Generates random passwords
+    .LINK
+        http://blog.simonw.se/powershell-generating-random-password-for-active-directory/
```
juhlee-microsoft: Instead of it, can you post the one from TechNet? When you are in this web page and find a link called Technet Galley. Could be safe if we link Technet url than external blog url.

juhlee-microsoft:
We wait for @lizzha update prior to merging.

Utilities/CreateServicePrincipal.ps1
lizzha: it was in the original code, I think I can remove this link since the TechNet link is on the top

lizzha:
@juhlee-microsoft  I've addressed the comments

PR #718, opened 2019-02-08
IB: Added reboot if IB is not active

adriansuhov:
An extra step was added in the Powershell script after all VMs have rebooted. A check for the modules will be made on each VM and if any ismissing, a reboot will be performed. We have 4 retries/VM. If after 4 retries IB is not up (modules are missing), then the test will be aborted.
e.g.
02/08/2019 14:37:39 : [INFO ] IB is initialized in client-vm-8
.....
02/08/2019 14:37:41 : [WARN ] IB is NOT initialized in client-vm-9
02/08/2019 14:37:42 : [INFO ] Triggering Restart-client-vm-9...
02/08/2019 14:37:42 : [INFO ] Waiting until VMs restart...
......

### DIFF
--- a/Testscripts/Windows/VERIFY-INFINIBAND-MultiVM.ps1
+++ b/Testscripts/Windows/VERIFY-INFINIBAND-MultiVM.ps1
@@ -1,6 +1,34 @@
 # Copyright (c) Microsoft Corporation. All rights reserved.
 # Licensed under the Apache License.
 
+function Resolve-UninitializedIB {
+    $cmd = "lsmod | grep -P '^(?=.*mlx5_ib)(?=.*rdma_cm)(?=.*rdma_ucm)(?=.*ib_ipoib)'"
+    foreach ($VmData in $AllVMData) {
+        $ibvOutput = ""
+        $retries = 0
+        while ($retries -lt 4) {
+            $ibvOutput = RunLinuxCmd -ip $VMData.PublicIP -port $VMData.SSHPort -username `
+                $test_super_user -password $password $cmd -ignoreLinuxExitCode:$true
+            if (-not $ibvOutput) {
+                LogWarn "IB is NOT initialized in $($VMData.RoleName)"
+                $restartStatus = RestartAllDeployments -AllVMData $VmData
+                Start-Sleep -s 20
+                if ($restartStatus -eq "True") {
+                    $retries++
+                } else {
+                    Throw "Failed to reboot $($VMData.RoleName)"
+                }
+            } else {
+                LogMsg "IB is initialized in $($VMData.RoleName)"
+                break
+            }
+        }
+        if ($retries -eq 4) {
+            Throw "After 4 reboots IB has NOT been initialized on $($VMData.RoleName)"
+        }
+    }
+}
+
 function Main {
     $resultArr = @()
     # Define two different users in run-time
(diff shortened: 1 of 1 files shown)


Testscripts/Windows/VERIFY-INFINIBAND-MultiVM.ps1
```diff
@@ -1,6 +1,34 @@
 # Copyright (c) Microsoft Corporation. All rights reserved.
 # Licensed under the Apache License.
 
+function Resolve-UninitializedIB {
+    $cmd = "lsmod | grep -P '^(?=.*mlx5_ib)(?=.*rdma_cm)(?=.*rdma_ucm)(?=.*rdma_ucm)'"
```
juhlee-microsoft: the last one should be ib_ipoib not rdma_ucm

Testscripts/Windows/VERIFY-INFINIBAND-MultiVM.ps1
```diff
@@ -1,6 +1,34 @@
...
+    foreach ($VmData in $AllVMData) {
+        $ibvOutput = ""
+        $retries = 0
+        while ($retries -lt 4) {
+            $ibvOutput = RunLinuxCmd -ip $VMData.PublicIP -port $VMData.SSHPort -username `
+                $test_super_user -password $password $cmd -ignoreLinuxExitCode:$true
+            if (-not $ibvOutput) {
+                LogWarn "IB is NOT initialized in $($VMData.RoleName)"
+                $restartStatus = RestartAllDeployments -AllVMData $VmData
+                Start-Sleep -s 20
```
juhlee-microsoft: Could be potential issue when VM size grows. Can you add here like another on you put in the code? Like waiting for VM state with timer?

Testscripts/Windows/VERIFY-INFINIBAND-MultiVM.ps1
```diff
@@ -160,7 +176,9 @@ function Main {
         $restartStatus = RestartAllDeployments -AllVMData $AllVMData
         LogMsg "Rebooting All VMs after all setup is done: $restartStatus"
         # Wait for VM boot up and update ip address
-        Start-Sleep -Seconds 240
+        Start-Sleep -Seconds 60
```
juhlee-microsoft: Why is it reduced to 25%?

Testscripts/Windows/VERIFY-INFINIBAND-MultiVM.ps1
adriansuhov: RestartAllDeployments function has all the wait/checks needed for the VMs to be fully booted up. This sleep is just an extra caution for IB only, maybe the related modules will take longer to load. Bottom line is that this is already tested and is working.

Testscripts/Windows/VERIFY-INFINIBAND-MultiVM.ps1
adriansuhov: ditto

Testscripts/Windows/VERIFY-INFINIBAND-MultiVM.ps1
adriansuhov: will change

adriansuhov:
Changed the code

PR #721, opened 2019-02-11
Update the default storage accounts

lizzha:
This should be included in the next stable release. Do not merge until we're about to release next stable.


juhlee-microsoft:
This turns out all pipeline job's configuration update, informing out to internal teams that use our storage account.  When PR is merging to master, all update/communication comes out also.

XML/RegionAndStorageAccounts.xml
```diff
@@ -1,111 +1,111 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <AllRegions>
     <eastasia>
-        <StandardStorage>eosgeastasia</StandardStorage>
-        <PremiumStorage>eosgeastasiax</PremiumStorage>
+        <StandardStorage>lisaeastasia</StandardStorage>
```
juhlee-microsoft: There are a few other files you should edit in this PR. Search up with 'eosg' and check out below files.
Azure.psm1
utils.sh
CAPTURE-VHD-BEFORE-TEST.ps1

lizzha: I double checked the 'eosg' occurrences in these files, they're used in resource names. The storage account change doesn't affect them.

chvalean:
@lizzha do we want to pick this up in the 1902 stable release?
Thinking that if we pick it in stable then we'll be forced to immediately check/fix pipelines or anything using it.
Or we release 1902 stable now, then merge this, which will give us additional time to handle any changes due to this?

lizzha:
@chvalean  do you mean release stable 1902 first, then merge this PR to stable branch? If so, I think it's OK.

chvalean:
> @chvalean do you mean release stable 1902 first, then merge this PR to stable branch? If so, I think it's OK.

Correct, this PR will be in the next stable - 1903 - next month, and in master branch until then.

lizzha:
@chvalean our goal is to migrate pipeline jobs to new subscription ASAP.
Are current pipeline jobs using master or stable? If they're using master, then we can exclude it from stable 1902. And later we merge the PR to master and update Jenkins at the same time.
If they're using stable, I think it's better to include this change in stable ASAP, and we update Jenkins at the same time when stable is released.

PR #745, opened 2019-02-14
Re-design VM size series

### LINKED ISSUE #645: Azure: VM size series needs better framework design
Upon new VM size added to the list of LISAv2, we used 'if-statement' with 'else-if'. Suggest to use "pattern" in the automation not continue on adding else-if statement.

LiliDeng:
resolve #645


Libraries/Azure.psm1
```diff
@@ -99,150 +99,116 @@ Function Validate-SubscriptionUsage($RGXMLData, $Location, $OverrideVMSize, $Sto
...
-                $currentStatus = Set-Usage -currentStatus $currentStatus -text $identifierText  -usage $testVMUsage -AllowedUsagePercentage $AllowedUsagePercentage
-                $overFlowErrors += Test-Usage -currentStatus $currentStatus -text $identifierText -AllowedUsagePercentage $AllowedUsagePercentage
-            }
-            #region M-Series
-            elseif ( $testVMSize.StartsWith("M")) {
-                $identifierText = "standardMSFamily"
-                $currentStatus = Set-Usage -currentStatus $currentStatus -text $identifierText  -usage $testVMUsage -AllowedUsagePercentage $AllowedUsagePercentage
-                $overFlowErrors += Test-Usage -currentStatus $currentStatus -text $identifierText -AllowedUsagePercentage $AllowedUsagePercentage
+            Switch ($testVMSize) {
+                {$_.StartsWith("DS") -and $_.EndsWith("v2")} {
```
lizzha: using regex is more concise

LiliDeng:
image

Libraries/Azure.psm1
LiliDeng: ok

chvalean:
@LiliDeng although this is a slight improvement from the original else/if loops, we still have to maintain a large list of VM sizes, so each time we want to add a new one, we introduce several lines of code.
Can you please research if there is a better way, wan't looked into this in detail but PowerShell should surely have some methods to define these better.

Even consider that the current list of VM sizes is not complete, so the entire list is much longer...

LiliDeng:
@chvalean Thanks, I will do some investigation.

Libraries/Azure.psm1
ader1990: I still do not like the if conditions, it seems that something can be better than this.

ifs can always be replaced by for loops with dictionaries, can you try this option?

ader1990: actually, this becomes simple if you do it like this:

$regExpVmSize = @{
  "standardDSv3Family" = @{ "Regexp" = "D*s_v3"; "isPremium"= 1};
}

foreach ($vmSize in $regExpVmSize.Keys) {
    if ($_ -match $regExpVMsize[$vmSize].RegExp) {
          $identifierTest = $vmSize
          $premiumVMs += $regExpVMsize[$vmSize].isPremium
          continue
    }
}

LiliDeng: make sense @ader1990

LiliDeng:
image

Libraries/Azure.psm1
```diff
@@ -131,160 +131,54 @@ Function Validate-SubscriptionUsage($RGXMLData, $Location, $OverrideVMSize, $Sto
...
+                "standardDSv3Family" = @{ "Regexp" = "^D.*s_v3$"; "isPremium"= 1};
+                "standardDSFamily" = @{ "Regexp" = "^DS((?!v2|s_v3).)*$"; "isPremium"= 1};
+                "standardDv2Family" = @{ "Regexp" = "^D[^S].*v2$"; "isPremium"= 0};
+                "standardDv3Family" = @{ "Regexp" = "^D[^S]((?!s_v3).)*v3$"; "isPremium"= 0};
+                "standardDFamily" = @{ "Regexp" = "^D[^S]((?!v2|v3).)*$"; "isPremium"= 0};
+                "standardESv3Family" = @{ "Regexp" = "^E.*s_v3$"; "isPremium"= 1};
+                "standardEv3Family" = @{ "Regexp" = "^E.*v3$"; "isPremium"= 0};
+                "standardA8_A11Family" = @{ "Regexp" = "A8|A9|A10|A11"; "isPremium"= 0};
+                "standardAv2Family"= @{ "Regexp" = "^A.*v2$"; "isPremium"= 0};
+                "standardA0_A7Family" = @{ "Regexp" = "^A((?!8|9|10|11|v2).)*$"; "isPremium"= 0};
```
lizzha: instead of excluding 8-11, why not specifying 0-7 directly?

PR #755, opened 2019-02-15
Fix typo for OL platform.

LiliDeng:


XML/TestCases/BVT.xml
```diff
@@ -308,7 +308,7 @@
         <testScript>BVT-VERIFY-DEPLOYMENT-PROVISION.ps1</testScript>
         <files></files>
         <setupType>OneVM</setupType>
-        <Platform>Azure,HyperV,OLQ</Platform>
+        <Platform>Azure,HyperV,OL</Platform>
```
juhlee-microsoft: There are more.

https://github.com/LIS/LISAv2/blob/bd7b22ecf715767c05d78b82eedd89d4d83b868e/LISAv2-Framework.psm1#L22
https://github.com/LIS/LISAv2/blob/bd7b22ecf715767c05d78b82eedd89d4d83b868e/LISAv2-Framework.psm1#L139
https://github.com/LIS/LISAv2/blob/bd7b22ecf715767c05d78b82eedd89d4d83b868e/Run-LisaV2.ps1#L35
https://github.com/LIS/LISAv2/blob/bd7b22ecf715767c05d78b82eedd89d4d83b868e/TestControllers/OLController.psm1#L28
https://github.com/LIS/LISAv2/blob/bd7b22ecf715767c05d78b82eedd89d4d83b868e/TestProviders/OLQProvider.psm1#L31

Last two lines are ok with as they-are, however 'OL' could be use as a part of function or variable name in other places. I suggest we should keep 'OL' only in all over the places.

juhlee-microsoft: @vyadavmsft Can you also post the test result with this change?

lizzha:
@LiliDeng it should be by design, there will be multiple platforms for OL, which share same test controller, but different test providers. @vyadavmsft should be working on this.

iamshital:
@LiliDeng , I believe, this was done on purpose. 

`OL refers to "Ov*****e"`
`OLQ refers to "Ov*****e on QEMU"`

@vyadavmsft Please confirm.

vyadavmsft:
discussed internally closing this.

PR #761, opened 2019-02-18
Enable nested virtualization in setup script, instead of in framework…

xian123:
… level


lizzha:
@xian123 please post the result for at least one case

Libraries/HyperV.psm1
```diff
@@ -407,29 +403,6 @@ function Create-HyperVGroupDeployment([string]$HyperVGroupName, $HyperVGroupXML,
     return $ReturnValue
 }
 
-function Enable-HyperVNestedVirtualization($HyperVGroupName, $HyperVHost) {
```
ader1990: you can leave this method here, and reuse it from the ps1 script

lizzha: @ader1990 why do you want to keep this method in HyperV.psm1? It's not used by other test besides nested cases on HyperV

ader1990: I was thinking that this method can be reused at some moment by other scripts. If this gets reused at some moment, here is the place it should be.

lizzha: no, it will only be used for nested VMs.

LiliDeng:
[LISAv2 Test Results Summary]
Test Run On           : 02/19/2019 08:11:07
VHD Under Test        : D:\lili\vhd\L1_win_2019.vhd
Total Test Cases      : 1 (1 Passed, 0 Failed, 0 Aborted, 0 Skipped)
Total Time (dd:hh:mm) : 0:1:0

    1 NESTED-HYPERV-NTTTCP-DIFFERENT-L1-PUBLIC-BRIDGE                                   PASS                 55.8

PR #762, opened 2019-02-18
Cleanup testcase xmls

asofron:
Some Test areas are  in lowercase in xmls , they should be only uppercase in order to be uniform . 
Also some testcase names need the  test area in the name to be easier to identify what the test case covers (some need to be in more uniform state )


chvalean:
LISAv2 is case-sensitive for the Area/Category/Testname.
Must notify all parties to ack the change and update pipelines and jobs.

PR #769, opened 2019-02-19
Fix: Replaced LogErr with Write-LogErr in SETUP-SRI-IOV-Configure

iamshital:
Also checked that "LogErr" is not present anywhere in LISAv2 powershell scripts.


iamshital:
A preventive action taken after https://github.com/LIS/LISAv2/pull/766

PR #785, opened 2019-02-21
HyperV.psm1: Added the hostname in /etc/hosts

adriansuhov:
Until now, the hostname was only injected in /etc/hostname. In some cases, if the hostname present in /etc/hosts and /etc/hostname don't match, an error (sudo: unable to resolve host) will pop every time
commands are run in sudo in VM. This is breaking a few tests.

## Before the fix:
2/21/2019 08:24:59 : [INFO ] The interface that will be configured on LISAv2-TwoVM1Dep-ubuntuAzuretrusty-SX79-636863052186-dependency-vm is
sudo: unable to resolve host (none)
eth1

## After the fix:
02/21/2019 10:40:48 : [INFO ] The interface that will be configured on LISAv2-TwoVM1Dep-ubAz-VG90-636863133854-dependency-vm is eth1


iamshital:
That's good. 👍

PR #790, opened 2019-02-21
Replace usages of tools folder to be case sensitive in powershell files

stefanlupsa:
the Tools folder is used with lowercase, this created a problem when I checked-out the repository using WSL and Powershell was no longer able to resolve the path case insensitive

### DIFF
--- a/Libraries/TestHelpers.psm1
+++ b/Libraries/TestHelpers.psm1
@@ -74,7 +74,7 @@ function Upload-RemoteFile($uploadTo, $port, $file, $username, $password, $usePr
 		if ($usePrivateKey)
 		{
 			Write-LogInfo "Uploading $file to $username : $uploadTo, port $port using PrivateKey authentication"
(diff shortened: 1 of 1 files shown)


juhlee-microsoft:
@stefanlupsa if this case-sensitivity does not work in WSL, this PR should fix all paths including test cases. There are 80+ lines with '\tools\' in repo. Please make all others. If you use a global variable, it would be better.

stefanlupsa:
> @stefanlupsa if this case-sensitivity does not work in WSL, this PR should fix all paths including test cases. There are 80+ lines with '\tools' in repo. Please make all others. If you use a global variable, it would be better.
 
@juhlee-microsoft 
I did a search for the tools path with lowercase uses in the PowerShell files over the whole repo, this is what I've found so far. I think we should merge this if it's good to go and I'll look for a more generalized solution to open a different PR

PR #852, opened 2019-03-06
Marked few LIS tests as SKIPPED instead of ABORTED if the kernel is not upgraded

iamshital:
## Description
In LIS deploy scenario tests, we have some tests where we first install the the latest kernel and then go for LIS installation.
Before this commit : 
If latest / minor kernel installation fails, LISAv2 marks the test as ABORTED.

After this commit : 
If latest / minor kernel installation fails, LISAv2 marks the test as SKIPPED.

---

### PR information
- [ ] The title of the PR is clear and informative.
- [ ] There are a small number of commits, each of which has an informative message. This means that previously merged commits do not appear in the history of the PR. For information on cleaning up the commits in your pull request, [see this page](https://github.com/Azure/azure-powershell/blob/master/documentation/development-docs/cleaning-up-commits.md).
- [ ] Except for special cases involving multiple contributors, the PR is started from a fork of the main repository, not a branch.
- [ ] If applicable, the PR references the bug/issue that it fixes in the description.
- [ ] Verify PR checker results and fix any coding errors.
- [ ] Included LISAv2 sample test results to demonstrate code functionality.

### Quality of Code and Contribution Guidelines
- [ ] I have read the [contribution guidelines](https://github.com/LIS/LISAv2/blob/master/.github/CONTRIBUTING.md).

### DIFF
--- a/Testscripts/Windows/LIS-INSTALL-SCENARIOS.ps1
+++ b/Testscripts/Windows/LIS-INSTALL-SCENARIOS.ps1
@@ -158,7 +158,7 @@ Function Uninstall-LIS ( $LISTarballUrlCurrent, $allVMData , $TestProvider) {
 Function Upgrade-Kernel ($allVMData, $TestProvider, [switch]$RestartAfterUpgrade){
     try {
         Write-LogInfo "Upgrading kernel"
-        Run-LinuxCmd -ip $allVMData.PublicIP -port $allVMData.SSHPort -username "root" -password $password -command "yum install -y kernel >> ~/kernel_install_scenario.log" -runMaxAllowedTime 20000
+        Run-LinuxCmd -ip $allVMData.PublicIP -port $allVMData.SSHPort -username "root" -password $password -command "yum install -y kernel >> ~/kernel_install_scenario.log" -runMaxAllowedTime 600 -maxRetryCount 3
         # Checking if latest kernel is already installed
         $sts = Run-LinuxCmd -ip $allVMData.PublicIP -port $allVMData.SSHPort -username "root" -password $password -command "cat kernel_install_scenario.log | grep 'already installed'" -ignoreLinuxExitCode:$true
         if ($sts) {
@@ -276,7 +276,7 @@ Function Install-LIS-Scenario-4 ($PreviousTestResult, $LISTarballUrlOld, $LISTar
     if ($PreviousTestResult -eq "PASS") {
         $UpgradekernelStatus=Upgrade-Kernel -allVMData $AllVMData -TestProvider $TestProvider
         if (-not $UpgradekernelStatus) {
-            return "FAIL"
+            return "SKIPPED"
         }
         $LISInstallStatus=Install-LIS -LISTarballUrl $LISTarballUrlCurrent -allVMData $AllVMData
         if ($LISInstallStatus -ne $false) {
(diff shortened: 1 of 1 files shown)


Testscripts/Windows/LIS-INSTALL-SCENARIOS.ps1
```diff
@@ -158,7 +158,7 @@ Function Uninstall-LIS ( $LISTarballUrlCurrent, $allVMData , $TestProvider) {
 Function Upgrade-Kernel ($allVMData, $TestProvider, [switch]$RestartAfterUpgrade){
     try {
         Write-LogInfo "Upgrading kernel"
-        Run-LinuxCmd -ip $allVMData.PublicIP -port $allVMData.SSHPort -username "root" -password $password -command "yum install -y kernel >> ~/kernel_install_scenario.log" -runMaxAllowedTime 20000
+        Run-LinuxCmd -ip $allVMData.PublicIP -port $allVMData.SSHPort -username "root" -password $password -command "yum install -y kernel >> ~/kernel_install_scenario.log" -runMaxAllowedTime 600 -maxRetryCount 3
```
chvalean: while at it... -username "root" - replace that with RunAsSudo instead?

Testscripts/Windows/LIS-INSTALL-SCENARIOS.ps1
```diff
@@ -276,7 +276,7 @@ Function Install-LIS-Scenario-4 ($PreviousTestResult, $LISTarballUrlOld, $LISTar
     if ($PreviousTestResult -eq "PASS") {
         $UpgradekernelStatus=Upgrade-Kernel -allVMData $AllVMData -TestProvider $TestProvider
         if (-not $UpgradekernelStatus) {
-            return "FAIL"
+            return "SKIPPED"
```
chvalean: I suppose this is for Azure where you might have the latest kernel already?
If so, then you should handle that better - I mean to handle the exact pattern that the running kernel is latest, and that would result in a Skip situation.
Anything else - like upgrading kernel actually failing, must still go as FAIL.

LiliDeng: @iamshital we should also skip other distro? LIS cases should only run against CentOS, RHEL and Oracle compatible with RHEL?

chvalean: > @iamshital we should also skip other distro? LIS cases should only run against CentOS, RHEL and Oracle compatible with RHEL?

That is handled at the test cases definition level, these tests have a separate Area and Tag.

Testscripts/Windows/LIS-INSTALL-SCENARIOS.ps1
iamshital: I will try with RunAsSudo. if it works, then good. If not, I will address in separate PR.

Testscripts/Windows/LIS-INSTALL-SCENARIOS.ps1
iamshital: @LiliDeng , These tests are part of  Functional -> LIS_DEPLOY area. And we execute it, only for supported distros.

iamshital: > I suppose this is for Azure where you might have the latest kernel already?
> If so, then you should handle that better - I mean to handle the exact pattern that the running kernel is latest, and that would result in a Skip situation.
> Anything else - like upgrading kernel actually failing, must still go as FAIL.

Yes. I'll have this tested for Azure and will do the necessary changes if any issue is found.

